PR #1, opened 2015-06-30
Install txt FILE from the HTML install file

ghost:
I have created an install txt file from the HTML version file.


highperformancecoder:
My first pull request! Thanks.

My main concern with this suggestion is that it creates two versions of the same thing, leading to a maintenance problem.

Whilst I'm personally in favour of simple text files for this information, others seem to like HTML.

What to do. One suggestion is to add a Makefile rule that builds the INSTALL.txt file if out of date with respect to the HTML file, eg 

lynx -dump INSTALL.html >INSTALL.txt

INSTALL.txt can still be a version controlled file so that it appears in github checked out code and releases - effectively it is a "Build file", in aegis terminology/

highperformancecoder:
Of course, it's not quite as simple as adding a Makefile rule, as the Makefiles are generated by qmake. You'll need to add a custom qmake target, or something. There's an example for how to do that with  tstarget in scidavis/basic.pri

ghost:
My idea was to have a standard file INSTALL in terms of GNU tools standard. Other standard file are:
NEWS, README, AUTHORS etc.

https://en.wikipedia.org/wiki/README

highperformancecoder:
I don't have a strong preference between plain text and html. See http://closingbraces.net/2007/08/07/readmetextorhtml/ for a discussion on pros and cons. But I strongly believe there shouldn't be both. So I'm inclined to leave things as they are. However, by all means raise the issue on scidavis-users or scidavis-contributers, and see what the feeling is.

fmeynadier:
Hi,

Isn't this exactly what Markdown is for ? for example if you rename the README file to README.md at the root of this repo, it will be nicely formatted in github while remaining also an easily editable text file.

(I see that wikipedia covers this particular usage... https://en.wikipedia.org/wiki/README#On_GitHub)

highperformancecoder:
That's a good suggestion. Markdown is a nice compromise between being
readable in a plain text editor and having nice mark up in a browser.

On Fri, Sep 04, 2015 at 05:36:48AM -0700, fmeynadier wrote:

> Hi,
> 
> Isn't this exactly what Markdown is for ? for example if you rename the README file to README.md at the root of this repo, it will be nicely formatted in github while remaining also an easily editable text file.
> 
> (I see that wikipedia covers this particular usage... https://en.wikipedia.org/wiki/README#On_GitHub)
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/highperformancecoder/scidavis/pull/1#issuecomment-137725367

## 

---

Prof Russell Standish                  Phone 0425 253119 (mobile)
Visiting Professor of Mathematics      hpcoder@hpcoders.com.au

## University of New South Wales          http://www.hpcoders.com.au

fmeynadier:
Ok, so it seems to work. I will propose the same kind of thing for INSTALL then...

PR #4, opened 2015-10-06
Applied https://sourceforge.net/p/liborigin/bugs/22/ to bundled liborigin

lupinix:
Applied https://sourceforge.net/p/liborigin/bugs/22/ to bundled liborigin


highperformancecoder:
I'm merging this, because %z appears to be C99 and C++-11 standard. If there is a problem, it will occur with Visual C++, but I suspect that compiler won't be used for compiling scidavis much.

PR #13, opened 2016-01-26
get rid of qtassistant & relace it with QDesktopServices

narunlifescience:
QDesktopServices is also supported in qt5.


narunlifescience:
I cant figure out why the test is throwing an exception. is that because there is no exception handling implemented ?

highperformancecoder:
On Tue, Jan 26, 2016 at 12:46:23PM -0800, Arun Narayanankutty wrote:

> I cant figure out why the test is throwing a exception. is that because there is no exception handling implemented ? 

It looks like showHelp is using a Critical message box to show the message
that the help files can't be found.

The regression suite turns all critical message boxes into exceptions,
as they are often used to indicate a critical failure of the software
instead of exceptions.

I suspect information might be a better level of message box to
use in showHelp. Alternatively, provide a dummy manual for the test
suite. Alternatively, find the real manual and place it in the source
code.

Someone sent me a copy of an updated version of the manual as a PDF,
as said he was writing it in LaTeX. I can follow up with him on that -
it seems a good idea forthe manual sources to be in LaTeX, and
converted to HTML for display at build time.

BTW - I've read up on QtDesktopServices, and it looks like exactly
what we need to replace QtAssistant with. Thanks for doing this.

Cheers

## 

---

## Economics, Kingston University         http://www.hpcoders.com.au

narunlifescience:
I changed critical QMessageBox to information. However I could not figure out how to fix error occurring at QDesktopServices::openUrl(QUrl(helpFilePath)). I did add index.html to manual folder but adding a dummy index file to the binary via .qrc files would not work with QDesktopServices::openUrl. How to fix this ?

Did you mean to add manual to scidavis releases rather than downloading from sourceforge ? converting Latex to html at build time could be done with some simple python script I guss.

highperformancecoder:
On Wed, Jan 27, 2016 at 12:04:31PM -0800, Arun Narayanankutty wrote:

> I changed critical QMessageBox to information. However I could not figure out how to fix error occurring at QDesktopServices::openUrl(QUrl(helpFilePath)). I did add index.html to manual folder but adding a dummy index file to the binary via .qrc files would not work with QDesktopServices::openUrl. How to fix this ?

Hmm - Looking at your code changes, I don't actually see why the test
is now failing when it was working previously.

There's a few possible answers:

1)  we don't unit test
ApplicationWindow::chooseHelpFolder() - not sure why it is being tested
anyway, but to disable that function call, add a function 

void ApplicationWindow::chooseHelpFolder() {}

to readWriteProject.cpp. This will be preferentially called over the
version in libscidavis.a, and effectively neuters that call for the
unit tests.

BTW - I don't like the fact that the whole unit test suite is in the file
readWriteProject, which is the name of the first unit test written. If
you have some thoughts about how to organise the unit tests, I'd love
to hear them.

2) Otherwise we need to add some sort of code to fire an event to the
dialog box being created, or otherwise indicate an alternative
location to a dummy index.html. This latter option has more appeal
when we actually add a manual into the repo.

For now, option 1) is probably the simplest.

> Did you mean to add manual to scidavis releases rather than downloading from sourceforge ? converting Latex to html at build time could be done with some simple python script I guss.

I usually use LaTeX2HTML, which is a not so simple perl script. The
other alternative is latexml, another perl script, which has some
advantages over LaTeX2HTML (works better for creating epubs, for
instance), but not others.

## Cheers

---

## Economics, Kingston University         http://www.hpcoders.com.au

narunlifescience:
I will try your suggestions..  
I am not familiar with Qtest class.. but I think we can split up tests based on different components or test types to different class for eg ReadWriteProjectTest::largeOriginImport() can be moved to another class lets say OriginTests along with some new unit tests. then use a central class lets say ScidavisTests and use objects to call functions from each classes. we can also move QTEST_MAIN to main.cpp if needed... I will check QTest documentation for more information..
I thought LaTeX to HTML conversion is simple but it seems not that simple indeed..
Thanks for reviewing the code..

narunlifescience:
BTW: I think we should use unique_ptr insted of auto_ptr.. auto_ptr seems to be deprecated as of c++ 11

highperformancecoder:
On Wed, Jan 27, 2016 at 09:02:34PM -0800, Arun Narayanankutty wrote:

> BTW: I think we should use unique_ptr insted of auto_ptr.. auto_ptr seems to be deprecated as of c++ 11

That begs the question of whether we should switch to coding in
C++11. At the moment, scidavis is C++98 compatible, but I've made the
switch on several other projects so it will not take much arm-twisting
to convince me we should change :).

And you're right - unique_ptr is what one should use for C++11.

Cheers

## 

---

## Economics, Kingston University         http://www.hpcoders.com.au

highperformancecoder:
On Wed, Jan 27, 2016 at 08:54:57PM -0800, Arun Narayanankutty wrote:

> I will try your suggestions..  
> I am not familiar with Qtest class.. but I think we can split up tests based on different components or test types to different class for eg ReadWriteProjectTest::largeOriginImport() can be moved to another class lets say OriginTests along with some new unit tests. then use a central class lets say ScidavisTests and use objects to call functions from each classes. we can also move QTEST_MAIN to main.cpp if needed... I will check QTest documentation for more information..

Nor me. But it appears to be the Qt annointed way of doing testing, so I'd
like to at least stick with it unless it becomes an obvious trouble.

> I thought LaTeX to HTML conversion is simple but it seems not that simple indeed..
> Thanks for reviewing the code..

## 

---

## Economics, Kingston University         http://www.hpcoders.com.au

narunlifescience:
had to add both showHelp() & chooseHelpFolder() to suppress the errors. Once we have the manual, we can add it to /manual folder. qmake is configured correctly to add this folder to /usr/share/doc/scidavis/ in case of linux/mac & applicationDirPath() for windows. 
I will look for a better solution later (as you rightly pointed out -> indicate an alternative location to a dummy index.html so that test suit can find it). We can just ignore it for now.. probably have more problems to deal with..

PR #15, opened 2016-01-28
add a custom IconHandler class

narunlifescience:
we should replace .xpm s with .svg and also use icons with different sizes rather than just one. This class would help to achive this in future. We should also try to use freedesktop icon names when possible.


narunlifescience:
OOPs the branch is pretty messed up. will close it for now.

PR #22, opened 2016-02-04
add a default table row height settings to preferences

narunlifescience:
The default raw height is fixed! This PR can be handy when we use scidavis with various display resolutions. here is the screenshot of added setting.. (default raw height range from 15 to 100)
table
here is the table view with default raw height set to 20:

here is the table view with default raw height set to 40:

### DIFF
--- a/libscidavis/src/ConfigDialog.cpp
+++ b/libscidavis/src/ConfigDialog.cpp
@@ -59,6 +59,7 @@
 #include <QVBoxLayout>
 #include <QListWidget>
 #include <QFontMetrics>
+#include <QSettings>
 
 ConfigDialog::ConfigDialog( QWidget* parent, Qt::WFlags fl )
     : QDialog( parent, fl )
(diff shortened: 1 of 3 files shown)


narunlifescience:
After having a look in to ConfigDialog class which deals with preferences, I find the implementation not that optimal. It uses public variables & slots from ApplicationWindow class to pass settings between them. The standard qt way would be to emit a signal & catch the signal from ApplicationWindow and then connect it to a private slot. this way, we can move a lot of public members to private & remove some unwanted public variables there by giving the code a more steam lined flow which could make it more elegant (in my perspective).. anothor thing is related to ui.. unless we are using highly customized qt widgets we can use .ui(xml files) for organizing widgets.. this would remove a significant amount of code to .ui xml files which is easy to manipulate & design using qt designer or creator.. the ui files will be converted to ui_c++ files employing User Interface Compiler (uic) at compile time.. 
These issues are not isolated ones but rather widespread through out scidavis code.. but to start with some thing simple like ConfigDialog would be easier..
what do you think ?

PR #32, opened 2016-05-31
+ added PRESET=new_qwt option, improved file locations

mianzhi:
PRESET=new_qwt option is added in the hope to solve the
canvas frame export issue by customizing the libqwt.

Not sure about other system, but the modified config.pri
works under ubuntu 16.04.

### DIFF
--- a/config.pri
+++ b/config.pri
@@ -35,6 +35,9 @@ unix:isEmpty(PRESET) { # allow command-line argument to override settings here
 	### Intended for Linux users building from source (on possibly somewhat
 	### older systems).
 	#PRESET = default_installation
+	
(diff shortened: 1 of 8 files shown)


highperformancecoder:
I don't want to merge these changes, as the builds will break on other systems.

My preferred approach to dealing with debian/ubuntu's directories is to create symbolic links

# hacky workaround of the fact the scidavis expects qwtplot3d to be in its own
# particular directory
	ln -sf /usr/include libscidavis/src/qwtplot3d
	ln -sf /usr/include/qwtplot3d-qt4 scidavis/qwtplot3d

But if you like, you could raise a discussion point on the mailing lists - maybe someone has a better way of dealing with the location.

PR #34, opened 2016-11-28
Fix wrong values on parameters' erros of fits - closes tickets #276/#290

Fellypao:
This simple commit fix the issues related with tickets #276 (Curve fit error estimates in SciDAVis D9) and #290( Wrong (and random) values of errors on fit parameters).


highperformancecoder:
Thanks Fellype. I'll check this out now.

PR #35, opened 2017-04-02
Liborigin singleversion

gbm19:
This pull request provides a new version of liborigin that replaces all the Origin??0Parser.{h,cpp} files
by a single file OriginAnyParser.{h,cpp} that can (hopefully) read all origin files from version 3.5 to current (2017) version.
It also includes the needed changes in the scidavis code to use the new version.
First commit has the largest change: it replaces the current liborigin (version 2.0.0) with the
code proposed at https://sourceforge.net/u/gbmiquel/liborigin/ci/singleversion/tree/
Implementation of that first commit can be revised at the circa 80 commits of the git branch above.
The other six commits are changes to scidavis to adapt it to the new liborigin.

The new version uses a different approach to parse origin files: reading the data is separated from interpreting it.
The main benefit of this method is that logging in "opjfile.log" will never grow to huge sizes as
it used to do when reading unsupported origin files.


highperformancecoder:
Hi gbm19 - sorry I've taken so long to get at this. Could you please supply some sample origin project files I can try this out on?

highperformancecoder:
Hi Miguel - I guess that's your real name. I just tried out your library on test/Histo.opj, but it doesn't load correctly. Is there any chance you could take a look at that example to see what's going wrong?

highperformancecoder:
Further to the above, it mostly loads correctly, but something goes awry with the plot titles.

gbm19:
I guess you mean that the "window title" of graph windows shows some "garbage"  when "Name and Label" is the window title mode. If so, please check my recent push. 
For origin project files to test I have been using example files in blog.originlab.com, or those at
http://users.physik.fu-berlin.de/~glaubitz/ (last link in "Debian Ubuntu"  section).

highperformancecoder:
Hi Miquel, any chance you could raise another pull request for your latest push? That way you get the credit :)

gbm19:
It is only a small change. I am going to push additional changes with (I
hope) improvements in opj parsing. Then I will issue another pull request.


2017-04-25 10:20 GMT+02:00 Russell Standish <notifications@github.com>:

PR #39, opened 2017-05-02
Three small improvements

gbm19:
This PR provides three small improvements. The most significant one is the speed up of large matrix data import.

### DIFF
--- a/libscidavis/src/importOPJ.cpp
+++ b/libscidavis/src/importOPJ.cpp
@@ -72,11 +72,14 @@ ImportOPJ::ImportOPJ(ApplicationWindow *app, const QString& filename) :
 {
 	xoffset=0;
 	try {
+		mw->setStatusBarText(QString("Import start ..."));
 		OriginFile opj((const char *)filename.toLocal8Bit());
 		parse_error = opj.parse();
+		mw->setStatusBarText(QString("... file parsed. Starting conversion to SciDAVis ..."));
 		importTables(opj);
 		importGraphs(opj);
 		importNotes(opj);
+		mw->setStatusBarText(QString());
 		if(filename.endsWith(".opj", Qt::CaseInsensitive))
 			createProjectTree(opj);
 		mw->showResults(opj.resultsLogString().c_str(), mw->logWindow->isVisible());
@@ -486,6 +489,7 @@ bool ImportOPJ::importTables(const OriginFile &opj)
 	int SciDAVis_scaling_factor=10; //in Origin width is measured in characters while in SciDAVis - pixels --- need to be accurate
 	for(unsigned int s = 0; s < opj.spreadCount(); ++s)
         {
+		mw->setStatusBarText(QString("Spreadsheet %1 / %2").arg(s+1).arg(opj.spreadCount()));
 		Origin::SpreadSheet spread = opj.spread(s);
 		int columnCount = spread.columns.size();
 		if(!columnCount) //remove tables without cols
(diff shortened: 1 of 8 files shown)


libscidavis/src/importOPJ.cpp
```diff
@@ -531,20 +531,12 @@ bool ImportOPJ::importTables(const OriginFile &opj)
...
 			Matrix->table()->blockSignals(true);
 #endif
-			unsigned int k=0;
-			for (int j=0; j<columnCount; j++)
-			{
-				for (int i=0; i<rowCount; i++)
-				{
-					double val = layer.data[k]; // matrix is the one in Origin file
-			        k++;
-					if(fabs(val)>0 && fabs(val)<2.0e-300)// empty entry
```
highperformancecoder, line 535: What was this guy smoking???

libscidavis/src/importOPJ.cpp
```diff
@@ -531,20 +531,12 @@ bool ImportOPJ::importTables(const OriginFile &opj)
...
-						continue;
-
-					Matrix->setCell(i, j, val); // Matrix is the one in Application
-					if (k >= layer.data.size()) break;
-				}
+			QVector<qreal> values;
+			values.resize(rowCount*columnCount);
+			for (int i=0; i<min((int)values.size(),(int)layer.data.size()); i++) {
+				values[i]=layer.data[i];
```
highperformancecoder, line 543: This seems an unnecessary copy to me. If setCells were reimplemented to just assign the first rowCount*columnCount entries of its arguments, then this assignment wouldn't be needed.

libscidavis/src/importOPJ.cpp
highperformancecoder: If these matrices really have non-existing entries, then they'd need to be marked by NaNs, in which case a test for finite() might be needed. What happens if a NaN is inserted into the matrix, as per your current code? Is SciDaVis happy? What happens if a particularly entry is not inserted into the matrix.

As it happens, if val is a NaN, then this conditional evaluates to false anyway. The smallest possible double precision IEEE number is 10^{-308}, but I think it is nuts to use specific genuine numbers to represent empty values, so I like your idea of getting rid of that functionality.

highperformancecoder (approved):

libscidavis/src/importOPJ.cpp
gbm19: Previous code had historical reasons. 
Origin had already in the early versions a special float value to represent NaN: the macro _ONAN at line 41 of OriginObj.h.

libscidavis/src/importOPJ.cpp
gbm19: But then you have to include OriginObj.h  in future_Matrix.{h,cpp}, which you probably do not want to.

libscidavis/src/importOPJ.cpp
highperformancecoder: Right - the only reason for doing that is because you need to support non-IEEE arithmetic. I can't remember the last time I used a machine that didn't use IEEE arithmetic, but I suspect it was during the 1980s. But the bizarre thing is that I don't think 10^-300 was even representable. Typically 8 bit exponents were used.

Anyway, I consider the use of 10^-300 to represent NaN as a bug, and so I'm happy it's now gone, but if you think we need bug-for-bug backward compatibility, then by all means put it back in (but use NaNs internally in SciDaVis if needed).

libscidavis/src/importOPJ.cpp
highperformancecoder: Why? SetCells knows how big its matrix is, and so does the layer.data object being passed. Just need for SetCells to set the minimum of both numbers, rather than bailing if the passed data does not match the matrix size.

PR #41, opened 2017-06-21
Remove boost dependency from liborigin

gbm19:
This PR provides the changes in liborigin and scidavis to remove boost (iequals, variant) from its build requires. The relevant changes in liborigin have been coded by Stefan Gerlach (liborigin, LabPlot).
I simply adapted them for use with scidavis.


highperformancecoder (approved):

highperformancecoder (approved):

highperformancecoder (approved):

3rdparty/liborigin/CMakeLists.txt
```diff
@@ -4,10 +4,6 @@ cmake_minimum_required(VERSION 2.8)
 # c++11 standard
 add_compile_options(-std=c++11)
 
-# boost library headers
-find_package(Boost REQUIRED)
```
highperformancecoder, line 8: Fine, but given we don't use cmake any more, this file can probably be removed in its entirety

3rdparty/liborigin/OriginObj.h
```diff
@@ -146,7 +169,76 @@ namespace Origin
...
+				break;
+			case V_STRING:
+				new(&as_string) string(v.as_string);
+				type = v.type;
+			}
+			return *this;
+		}
+*/
+		~Variant() {
+	//		printf("~Variant()\n");
```
highperformancecoder: Leaving the destructor unimplemented will cause memory leaks whenever a string is stored in this type.

3rdparty/liborigin/OriginObj.h
```diff
@@ -146,7 +169,76 @@ namespace Origin
...
+
+			if (type == V_STRING)
+				as_string.~string();
+
+			switch (v.type) {
+			case V_DOUBLE:
+				as_double = v.as_double;
+				break;
+			case V_STRING:
+				new(&as_string) string(v.as_string);
```
highperformancecoder: This can be replaced by assignment, and the destructor on line 223 removed.

3rdparty/liborigin/opj2dat.cpp
```diff
@@ -98,18 +98,18 @@ int main(int argc, char *argv[]) {
...
 							if (v != _ONAN) {
 								outf << v << "; ";
 							} else {
 								outf << nan("NaN") << "; ";
 							}
 						}
-						if (value.type() == typeid(string)) {
-							outf << boost::get<string>(value).c_str() << "; ";
+						if (value.type == Origin::variant::V_STRING) {
+							outf << (value.as_string).c_str() << "; ";
```
highperformancecoder, line 112: c_str() is redundant here.

highperformancecoder (changes requested):

highperformancecoder:
Please fix the memory leak in the implementation of Origin::variant.

gbm19:
Well, someday there will be a new liborigin release. Then it will be
treated like the other libs (e.g., muparser).
So I prefer to keep the CMakeList.txt, even though liborigin is so simple
now that you can  almost compile it by hand.
Also, I believe that Scidavis would benefit from switching to cmake instead
of qmake for compiling.

2017-06-22 1:02 GMT+02:00 Russell Standish <notifications@github.com>:

highperformancecoder:
If/when liborigin is restored as a standalone project, do you think this code is likely to make its way into that? Just wondering, because I was contemplating moving the liborigin code back into the scidavis codebase, and giving up on the pretense that it is a vendor branch.

Re qmake vs cmake, I personally think that both options are equally dreadful. But I do agree with (I think it was Dmitriy), that having chosen one system, we should stick with it consistently.

highperformancecoder (approved):

highperformancecoder:
The Travis failure looks unrelated - perhaps due to the recent upgrade of system images on Travis.

PR #45, opened 2017-07-20
Add python methods to modify attached axis of a curve

gbm19:


highperformancecoder:
Can't we use enums directly, rather than integer equivalents, for more readable code?

PR #47, opened 2017-08-06
Add the converted man page and install it

Mailaender:
I tried to compile it as instructed, but that failed assuming due to bit rot or Debian specific formats.

```
[  215s] + docbook-to-man man/manpage.sgml
[  215s] man/manpage.sgml:1: parser error : StartTag: invalid element name
[  215s] <!doctype refentry PUBLIC "-//OASIS//DTD DocBook V4.1//EN" [
[  215s]  ^
[  215s] man/manpage.sgml:1: parser error : Extra content at the end of the document
[  215s] <!doctype refentry PUBLIC "-//OASIS//DTD DocBook V4.1//EN" [
[  215s]  ^
[  215s] unable to parse man/manpage.sgml
[  215s] Unable to recognise encoding of this document at /usr/lib/perl5/vendor_perl/5.24.1/XML/SAX/PurePerl/EncodingDetect.pm line 100.
[  215s] Document requires an element [Ln: 1, Col: 0]
```

As on @openSUSE the `scidavis.1` was supplied as an additional source and it is also a text file though a way simpler one, I decided to replace it here, so every distribution can ship it from now on with no extra effort. `docbook-to-man` also took a lot of dependencies to install which simplifies the build.


highperformancecoder:
I was actually unaware SciDaVis came with a man page, so thanks for this. Looking over the man page, I suspect the --manual option won't work, and the --execute option might have changed slightly.

PR #54, opened 2017-09-20
Add calculation and display of curve integral

gbm19:
This is a first attempt to implement feature request #100.

### DIFF
--- a/libscidavis/src/Filter.cpp
+++ b/libscidavis/src/Filter.cpp
@@ -235,6 +235,8 @@ void Filter::output()
     calculateOutputData(X, Y);
 
 	addResultCurve(X, Y);
+	delete[] X;
+	delete[] Y;
 }
 
 int Filter::sortedCurveData(QwtPlotCurve *c, double start, double end, double **x, double **y)
@@ -321,8 +323,6 @@ QwtPlotCurve* Filter::addResultCurve(double *x, double *y)
 	d_graph->insertPlotItem(c, Graph::Line);
     d_graph->updatePlot();
 
-    delete[] x;
-	delete[] y;
 	return (QwtPlotCurve*)c;
 }
 
--- a/libscidavis/src/Integration.cpp
+++ b/libscidavis/src/Integration.cpp
@@ -69,6 +69,27 @@ void Integration::init()
 	d_result = NAN;
 }
 
+double Integration::trapezoid()
+{
+	double sum = 0.0;
+	double *result = (double *)malloc(d_n*sizeof(double));
+	int size = d_n - 1;
+	for(int i=0; i < size; i++){
+		int j = i + 1;
+		if (result)
+            result[i] = sum;
+		sum += 0.5*(d_y[j] + d_y[i])*(d_x[j] - d_x[i]);
+	}
+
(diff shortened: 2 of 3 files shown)


highperformancecoder (approved):
My main comment is that we should be adding RAII code for any new work to improve exception safety. In this case, a vector<double> might be a suitable replacement for new double[], and malloc below. If you are concerned about paying unneeded initialisation costs, there are ways of writing the code to avoid it, using reserve() and emplace_back(). Alternatively, use a roll-your-own container (eg ecolab::array) that does not initialise its members

I know scidavis is not currently exception safe, but we should aim for it ultimately.

gbm19:
Please check attached patch to improve Integration::trapezoid().
Replacing 'new' and 'delete' in Filter::output() and others in Filter.cpp
would require rewriting all files that include Filter.h
[use-gsl-vectors-in-trapezoid.patch.txt](https://github.com/highperformancecoder/scidavis/files/1320894/use-gsl-vectors-in-trapezoid.patch.txt)

highperformancecoder:
Why is this an improvement?

gbm19:
Because I believe gsl library is exception safe (I can be wrong). 
Even if it is not, gsl was already used in Integration.cpp, so if in the future gsl is replaced by something better, it will also be replaced in trapezoid().

gbm19:
Maybe you  prefer this
[use-std-vector-in-trapezoid.patch.txt](https://github.com/highperformancecoder/scidavis/files/1323955/use-std-vector-in-trapezoid.patch.txt)

highperformancecoder:
On Fri, Sep 22, 2017 at 08:52:55AM +0000, gbm19 wrote:
> Maybe you  prefer this
> [use-std-vector-in-trapezoid.patch.txt](https://github.com/highperformancecoder/scidavis/files/1323955/use-std-vector-in-trapezoid.patch.txt)
> 
> 
> -- 
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub:
> https://github.com/highperformancecoder/scidavis/pull/54#issuecomment-331390485

<professor mode on>
Yes, that version is RAII, the GSL version is not. Did you look up the
concept of RAII and the issue of exception safety? RAII also makes it
very difficult to code memory leaks.

BTW GSL is a C language API, so has no inbuilt concept of exceptions,
let alone exception safety. Were you confusing it with thread safety?
</professor mode off>

Cheers
-- 

gbm19:
Well, I was mislead by a too quick look at the results of googling gsl+RAII.
Thanks for clarifying.
Please, apply the vector version patch.

highperformancecoder:
On Fri, Sep 22, 2017 at 12:18:56PM +0000, gbm19 wrote:
> Well, I was mislead by a too quick look at the results of googling gsl+RAII.
> Thanks for clarifying.
> Please, apply the vector version patch.


Sure, well will do. Of course, I could always do that, but code review
does give educational opportunities too.

RAII is probably the most important C++ concept. After that, const
correctness is also very important. SciDaVis has been mostly written
without cognizance of these basic principles, so there's a lot of work
to bring the codebase up to spec.


-- 

PR #66, opened 2018-03-08
Improve speed of Paste operation with large data sets

gbm19:
Fix for bug #292

### DIFF
--- a/libscidavis/src/future/table/TableView.cpp
+++ b/libscidavis/src/future/table/TableView.cpp
@@ -799,6 +799,28 @@ IntervalAttribute<bool> TableView::selectedRows(bool full) {
 	return result;
 }
 
+bool TableView::hasMultiSelection()
+{
+	QModelIndexList indexes = d_view_widget->selectionModel()->selectedIndexes();
+	if (indexes.size() < 2)
+		return false;
+
+	QModelIndex index;
+	int minrow = indexes.at(0).row();
+	int maxrow = minrow;
+	int mincol = indexes.at(0).column();
+	int maxcol = mincol;
+	foreach (index, indexes)
+	{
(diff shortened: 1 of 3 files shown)


highperformancecoder (commented):
I'm not sure this is quite equivalent functionality. If my target selection were smaller than the text being pasted, current functionality is discard the clipboard data that doesn't fit. However, this new code appears to paste the selection regardless of the selection's size.

Fellypao:
A non-technical comment :-)

> If my target selection were smaller than the text being pasted, current functionality is discard the clipboard data that doesn't fit.

It will occur **only** If you select the columns before pasting.
If you just click on the **first cell** of a table, the selection will be extended to the clipboard data dimensions.
If you click on **any single cell** of a table, say in the ij position, then the selection will be extended **from** the ij position to the clipboard data dimensions.
All these things are the intended behavior according to the former SciDAVis developers (I remember that I've commented this to you some time ago...). Of course you can change this behavior to a more intuitive one, if you want.

On simple tests, using data from 4 columns and 20 rows from LibreOffice Calc:
1: selecting the first cell of a table:
- pasting the data on a default SciDAVis table (2 columns vs 30 rows) the result was a table with 4 columns and 30 rows
- pasting the data on a 1x1 table the result was a table with 4 columns and 20 rows

2: selecting the cell 2,2 of a 2x2 table: the result was a table with 5 columns and 21 rows.

gbm19:
This commit adds to code to keep current functionality.

highperformancecoder:
Thanks Miquel for your stellar efforts. @Fellypao , I agree that pasting the full block should be the behaviour when a single cell target selection is made. It's a good thing that was the previous behaviour, and now with this pull request it remains the behaviour.

PR #67, opened 2018-03-29
Dropqt3

gbm19:
Remove the qt3support dependency. All qt3 code has been replaced by equivalent qt4 code.


highperformancecoder:
Its failing one of the python tests ATM. I'll merge this into a branch and debug the issue when I get back to Sydney next week.

highperformancecoder:
I have merged this into a dropqt3 branch. Regression tests appear to be fixed - maybe your last commit fixed things. As mentioned, next I want to create regression tests covering the lines you've touched, and if there's no issues, I'll merge up to master.

libscidavis/src/AxesDialog.cpp
```diff
@@ -1428,11 +1428,11 @@ void AxesDialog::updateScale()
 	int a = Graph::mapToQwtAxis(axis);
 	const QwtScaleDiv *scDiv=d_plot->axisScaleDiv(a);
 #if QWT_VERSION >= 0x050200
-	boxStart->setText(QString::number(QMIN(scDiv->lowerBound(), scDiv->upperBound())));
-	boxEnd->setText(QString::number(QMAX(scDiv->lowerBound(), scDiv->upperBound())));
+	boxStart->setText(QString::number(qMin(scDiv->lowerBound(), scDiv->upperBound())));
```
highperformancecoder, line 1431: My preference would be to use std::min & std::max functions here, rather than the qt specific variants. std::min, max has been in the standard since C++98.

libscidavis/src/ApplicationWindow.cpp
```diff
@@ -7358,7 +7358,7 @@ void ApplicationWindow::activateWindow(MyWidget *w)
 
 void ApplicationWindow::maximizeWindow(QTreeWidgetItem * lbi)
 {
-	if (!lbi || lbi->rtti() == FolderListItem::RTTI)
+	if (!lbi || lbi->type() == FolderListItem::FolderType)
```
highperformancecoder: The better way to do rtti is not to have a type attribute, but rather do a reinterpret_cast to a pointer of the appropriate type. Eg in this example, the if condition can be written:
~~~~~
if (auto wli=dynamic_cast<WindowListItem*>(lbi))
  if (auto w=wli->window())
     {
	w->setMaximized();
 	updateWindowLists(w);
        emit modified();
     }
~~~~~
much cleaner, and more correct

gbm19:
My goal was to minimize the changes when removing the qt3support dependency.
Moreover, there are additional uses of qMin/qMax in scidavis and qwt-5.2.3
code. And in qwt-6.x code also.
If qMin/qMax has to be replaced by std::min/max, then it should be done
everywhere.



2018-04-23 3:31 GMT+02:00 Russell Standish <notifications@github.com>:

gbm19:
Maybe. Again, my goal was to minimize changes when removing qt3support.
Also, Qt4/Qt5 QTreeWidgetItem already has the type attribute, whether we
use it or not.
Using dynamic_cast can be done in next refactoring step, as there are other
uses of rtti in the code.

2018-04-23 6:24 GMT+02:00 Russell Standish <notifications@github.com>:

highperformancecoder:
On Mon, Apr 23, 2018 at 09:54:06AM -0700, gbm19 wrote:
> Maybe. Again, my goal was to minimize changes when removing qt3support.
> Also, Qt4/Qt5 QTreeWidgetItem already has the type attribute, whether we
> use it or not.
> Using dynamic_cast can be done in next refactoring step, as there are other
> uses of rtti in the code.
> 

I wasn't suggesting a "war on rtti". That's probably a huge waste of
time. Just that when you're refactoring a bit of code, you may as well
write it in modern idiom, depending on just how much work is involved,
of course. 

Yes - it is unfortunate that Qt invents its own way of doing things
that are already covered by elements of the standard language/standard
library. That may well be for well understandable historical
reasons. My attitude is that the standard is better known than Qt (or
should be) such that doing things the standard way is to be preferred.


-- 

gbm19:
Last two commit are also suitable for the master branch.

PR #86, opened 2018-07-05
Set added columns as Numeric instead of Text

gbm19:
This PR fixes the second issue in [https://sourceforge.net/p/scidavis/mailman/message/36262026/]
(scidavis-user mailing list)

### DIFF
--- a/libscidavis/src/future/table/future_Table.cpp
+++ b/libscidavis/src/future/table/future_Table.cpp
@@ -450,7 +450,7 @@ void Table::pasteIntoSelection()
 				QList<Column*> cols;
 				for(int i=0; i<last_col+1-columnCount(); i++)
 				{
-					Column * new_col = new Column(QString::number(i+1), SciDAVis::Text);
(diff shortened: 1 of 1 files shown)


libscidavis/src/future/table/future_Table.cpp
```diff
@@ -450,7 +450,7 @@ void Table::pasteIntoSelection()
 				QList<Column*> cols;
 				for(int i=0; i<last_col+1-columnCount(); i++)
 				{
-					Column * new_col = new Column(QString::number(i+1), SciDAVis::Text);
+					Column * new_col = new Column(QString::number(i+1), SciDAVis::Numeric);
```
highperformancecoder, line 453: I guess it make sense to assume that pasted data will be numeric, but what happens if it contains dates, formulae or strings? With your change, the pasted data will be lost, as opposed to the user having to convert the new columns manually. I'm going to merge this PR, but raise a ticket on the best way to handle arbitrary data.

PR #87, opened 2018-07-14
Fix bug #341: Pasting data into table cells loses data ...

gbm19:


libscidavis/src/future/table/future_Table.cpp
```diff
@@ -497,9 +497,15 @@ void Table::pasteIntoSelection()
 				cols_texts << cur_column;
 			}
 
+			// regular expression for numeric data
+			QRegExp floatorintrx = QRegExp("^(([-+]?\\d*\\.\\d+([eE][-+]?\\d+)?)|([-+]?\\d+([eE][-+]?\\d+)?))$");
```
highperformancecoder: I'm pretty sure it is acceptable for numbers to have no digits trailing the decimal point, ie 1.e6 is accepted if you type it into a SciDAVis table cell.

This regular expression will reject such numbers. Otherwise, this PR looks good.

libscidavis/src/future/table/future_Table.cpp
```diff
@@ -498,12 +498,12 @@ void Table::pasteIntoSelection()
 			}
 
 			// regular expression for numeric data
-			QRegExp floatorintrx = QRegExp("^(([-+]?\\d*\\.\\d+([eE][-+]?\\d+)?)|([-+]?\\d+([eE][-+]?\\d+)?))$");
+			QRegExp floatorintrx = QRegExp("^[-+]?(?=.*\\d)\\d*\\.?\\d*([eE][-+]?\\d+)?$");
```
highperformancecoder: Well, you stretched my knowledge of Regexps here! It took me a while to figure out what the lookahead expression was doing, but I gather it is to reject patterns not containing a digit, eg "+." which would otherwise match the remainder of the RegExp.

PR #88, opened 2018-07-17
Fix bug #341: continued.

gbm19:
The regex still had a flaw. 
My initial idea was to prevent '.' match the regex, but after your comment I realised that '+.E3' matches and should not.
Hence, a new regex is needed.

### DIFF
--- a/libscidavis/src/future/table/future_Table.cpp
+++ b/libscidavis/src/future/table/future_Table.cpp
@@ -498,7 +498,7 @@ void Table::pasteIntoSelection()
 			}
 
 			// regular expression for numeric data
-			QRegExp floatorintrx = QRegExp("^[-+]?(?=.*\\d)\\d*\\.?\\d*([eE][-+]?\\d+)?$");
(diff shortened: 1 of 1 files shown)


highperformancecoder:
On Tue, Jul 17, 2018 at 10:03:58AM -0700, gbm19 wrote:
> The regex still had a flaw. 
> My initial idea was to prevent '.' match the regex, but after your comment I realised that '+.E3' matches and should not.
> Hence, a new regex is needed.

Good point! 


-- 

libscidavis/src/future/table/future_Table.cpp
```diff
@@ -498,7 +498,7 @@ void Table::pasteIntoSelection()
 			}
 
 			// regular expression for numeric data
-			QRegExp floatorintrx = QRegExp("^[-+]?(?=.*\\d)\\d*\\.?\\d*([eE][-+]?\\d+)?$");
+			QRegExp floatorintrx = QRegExp("^[-+]?(?=[^eE]*\\d)\\d*\\.?\\d*([eE][-+]?\\d+)?$");
```
highperformancecoder, line 501: I believe you have done it, sir!

PR #96, opened 2018-09-10
Qt4 to Qt5, third step

gbm19:
This code allows to build SciDAVis with either Qt5 or Qt4.
Qt version is chosen by using the corresponding qmake binary.
For the Qt5 version a port of qwtplot3d and qwt5 to Qt5 is needed.
Fedora and Debian already have a package for qwtplot3d-qt5.
For qwt5-qt5 you can get and compile the code at https://github.com/gbm19/qwt5-qt5
The pull request also contains some small bug fixes unrelated to Qt5.

### DIFF
--- a/libscidavis/libscidavis.pro
+++ b/libscidavis/libscidavis.pro
@@ -1,9 +1,7 @@
 # enable C++11 support
-greaterThan(QT_MAJOR_VERSION, 3){
-  CONFIG += c++11
-} else {
-  QMAKE_CXXFLAGS += -std=c++0x
-}
+equals(QT_MAJOR_VERSION, 5) { CONFIG += c++11 }
+equals(QT_MAJOR_VERSION, 4) { QMAKE_CXXFLAGS += -std=c++11 }
+equals(QT_MAJOR_VERSION, 3) { QMAKE_CXXFLAGS += -std=c++0x }
 
 TEMPLATE=lib
 CONFIG+=staticlib uic
--- a/libscidavis/libscidavis.pro
+++ b/libscidavis/libscidavis.pro
@@ -1,5 +1,5 @@
 # enable C++11 support
-greaterThan(QT_MAJOR_VERSION, 4){
+greaterThan(QT_MAJOR_VERSION, 3){
   CONFIG += c++11
 } else {
   QMAKE_CXXFLAGS += -std=c++0x
@@ -8,7 +8,7 @@ greaterThan(QT_MAJOR_VERSION, 4){
 TEMPLATE=lib
 CONFIG+=staticlib uic
 TARGET=scidavis
-QMAKE_CLEAN+=${TARGET}
+QMAKE_CLEAN += $$TARGET
 
 include(../config.pri)
 
--- a/libscidavis/libscidavis.pro
+++ b/libscidavis/libscidavis.pro
@@ -31,7 +31,9 @@ DEFINES       += MANUAL_PATH="\\\"$$replace(manual.path," ","\\ ")\\\""
 !mxe {
      win32:DEFINES += QT_DLL QT_THREAD_SUPPORT
 }
+
 QT            += opengl network svg xml
+equals(QT_MAJOR_VERSION, 5) { QT += printsupport }
 
 MOC_DIR        = ../tmp/scidavis
 OBJECTS_DIR    = ../tmp/libscidavis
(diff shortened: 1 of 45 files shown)


libscidavis/src/Graph.cpp
```diff
@@ -1399,7 +1399,16 @@ void Graph::exportVector(const QString& fileName, int, bool color, bool keepAspe
...
-    	printer.setOutputFormat(QPrinter::PostScriptFormat);
+    {
+#if QT_VERSION >= 0x050000
+        QMessageBox::warning(this, tr("Warning"),
+            tr("Output in postscript format is not available for Qt5, using PDF"));
+        printer.setOutputFormat(QPrinter::PdfFormat);
+        printer.setOutputFileName(fileName+".pdf");
+#else
+        printer.setOutputFormat(QPrinter::PostScriptFormat);
+#endif
```
highperformancecoder, line 1410: So QT5 dropped postscript support? I guess there are tools for converting PDF to postscript, so this is not fatal, but really? People still use EPS extensively for incorporating graphics.

libscidavis/libscidavis.pro
```diff
@@ -1,9 +1,7 @@
 # enable C++11 support
-greaterThan(QT_MAJOR_VERSION, 3){
-  CONFIG += c++11
-} else {
-  QMAKE_CXXFLAGS += -std=c++0x
-}
+equals(QT_MAJOR_VERSION, 5) { CONFIG += c++11 }
+equals(QT_MAJOR_VERSION, 4) { QMAKE_CXXFLAGS += -std=c++11 }
+equals(QT_MAJOR_VERSION, 3) { QMAKE_CXXFLAGS += -std=c++0x }
```
highperformancecoder, line 4: Why does this depend on Qt version? This is a compiler flag, so I would have though it'd depend on compiler version only.

libscidavis/src/Graph.cpp
gbm19: Nowadays SVG is for me a much better vector format than postscript.
And as you say there are tools to convert to postscript from PDF/SVG.
Depending on the user's reaction, we can improve current solution.

libscidavis/libscidavis.pro
gbm19: Yes, but how you tell qmake the relevant compiler flags depends on the Qt version.
For Qt5 this is better done through the CONFIG variable (http://doc.qt.io/qt-5/qmake-variable-reference.html)
For Qt3/Qt4 the only way was by extending the QMAKE_CXXFLAGS variable.

highperformancecoder:
On Thu, 13 Sep 2018, 04:29 gbm19, <notifications@github.com> wrote:

> *@gbm19* commented on this pull request.
> ------------------------------
>
> In libscidavis/libscidavis.pro
> <https://github.com/highperformancecoder/scidavis/pull/96#discussion_r217177187>
> :
>
> > @@ -1,9 +1,7 @@
>  # enable C++11 support
> -greaterThan(QT_MAJOR_VERSION, 3){
> -  CONFIG += c++11
> -} else {
> -  QMAKE_CXXFLAGS += -std=c++0x
> -}
> +equals(QT_MAJOR_VERSION, 5) { CONFIG += c++11 }
> +equals(QT_MAJOR_VERSION, 4) { QMAKE_CXXFLAGS += -std=c++11 }
> +equals(QT_MAJOR_VERSION, 3) { QMAKE_CXXFLAGS += -std=c++0x }
>
> Yes, but how you tell qmake the relevant compiler flags depends on the Qt
> version.
> For Qt5 this is better done through the CONFIG variable (
> http://doc.qt.io/qt-5/qmake-variable-reference.html)
> For Qt3/Qt4 the only way was by extending the QMAKE_CXXFLAGS variable.
>
OK - I wasn't concerned about that, but having different -std flags. Let's
use -std=c++11 everywhere, and get rid of std=c++0x

What are the advantages of using CONFIG+=c++ over just setting the compiler
flag?

gbm19:
> OK - I wasn't concerned about that, but having different -std flags. Let's
> use -std=c++11 everywhere, and get rid of std=c++0x
>
> What are the advantages of using CONFIG+=c++ over just setting the compiler
> flag?
>

The main one is that one can use 'c++11' in the same way the other keys in
CONFIG are used: to conditionally include declarations.
See http://doc.qt.io/qt-5/qmake-language.html#scopes, paragraf "Section and
scopes".

Also using 'c++11' you don't need to find out the specific switches for
each compiler you want to use, qmake takes care of it.
Unfortunately, Qt4 does not accept the 'c++11' value for CONFIG as I
learned the hard way with this PR when travis used the c++98 standard to
compile.
Thus, if we ignore Qt3, we could have:

>
greaterThan(QT_MAJOR_VERSION, 4) {
> CONFIG += c++11
> } else {
> QMAKE_CXXFLAGS += -std=c++11
> }
>

And when working only with Qt5:

> CONFIG += c++11
>

Issue #64, opened 2018-02-14
not enough colors


yeKcim:
there is a lack of possibility for lines color (in graph)

Sometimes (just as example), you've got 6 curves to display (3 theoretical and 3 measured), colors could be used to mean something (red gradient for theoretical, blue one for measured). There is a lot of examples where colors mean something.

The choice is actually too limited, I had to draw some graphs with pyplot instead of scidavis 💔

yeKcim:
I've just remember that I opened https://sourceforge.net/p/scidavis/scidavis-bugs/278/ previously

> Please change the color picker for a modern one, like https://github.com/mbasaglia/Qt-Color-Widgets/blob/master/gallery/README.md for example.
> 
> For curve color choice, let the user select more than 24 colors, it's impossible to draw a graph with 4 curves that use 4 variations of orange, for example. It could be a big problem

![capture du 2017-10-18 14-06-47](https://user-images.githubusercontent.com/10558744/36205797-42c9e7da-1190-11e8-8d18-db7e8d9030e7.png)

Fellypao:
Hi @yeKcim ,
I agree with you that SciDAVis could have more colors available for curves. Currently, only the error bars have a rich variety of colors:
![colors](https://user-images.githubusercontent.com/16227753/36565889-984c4f24-1800-11e8-8975-a811f68afae7.png)

yeKcim:
Is that difficult to always use the same powerfull color-picker in the application?

ghost:
Hello,
I just hacked the colour picker into the line options. @yeKcim, would the image below represent what you imagined? Also, do you know of more color lists that should be replaced with a picker?

![scidavis_color_picker](https://user-images.githubusercontent.com/26203394/38172076-14d3ec6c-3595-11e8-8f52-ef4932f7dad3.png)

yeKcim:
really good! Thanks a lot kadabash
I also see "Fill color" in the same window, and colors in sectors properties.

Fellypao:
Hello @Kadabash

> I just hacked the colour picker into the line options

This looks great. Good job.

>  Also, do you know of more color lists that should be replaced with a picker?

In principle, all color lists in dialog windows could be replaced by a picker:
1) in the Analysis menu:
- Smooth -> all smooth options
- Same for FFT filter, Interpolate, FFT...
- Some fit options under Quick Fit and also in the Fit Wizard

2) in the General Plot Options:
- Grid -> Major and Minor grids

[...]

PR #104, opened 2018-10-09
Color enhance

gbm19:
This PR systematizes the use of ColorButton to define all colors in graphics.
It also allows to choose an alpha value (transparency/opacity) for each color.
Color values (including alpha) are stored in .sciproj files as #AARRGGBB.
Thus, files saved with this version can not be properly read with previous SciDAVis versions that stored colors as #RRGGBB.
It should solve github issue #64, and sf #278.


libscidavis/src/ColorButton.cpp
```diff
@@ -28,18 +28,55 @@
...
 
 void ColorButton::init()
 {
+	for (int i=0; i<8; i++)
+	{
+		for (int j=0; j<3; j++)
+		{
+			QColorDialog::setStandardColor(6*i+j,   colors[i+8*j].rgb());
+		}
+	}
```
highperformancecoder, line 107: The magic numbers 8 and 3 bother me a bit. I assume that maybe the dialog is perhaps 8x3 or something? At very least, there should be a static_assert that 8*3 <= colors_count.

Also what happens to standard colours 3,4,5, 9,10,11, ...?

libscidavis/src/ColorButton.cpp
```diff
@@ -61,22 +98,65 @@ void ColorButton::init()
...
 }
 
+unsigned int ColorButton::colorIndex(const QColor& c)
+{
+	const QColor *ite = std::find(std::begin(colors), std::end(colors), c);
+	if (ite->isValid())
+		return (ite - colors);
+	else
+		return c.rgba();
+}
```
highperformancecoder, line 152: Kind of a nice hack, but what happens when someone wants a very translucent shade of black?

libscidavis/src/ColorButton.cpp
```diff
@@ -61,22 +98,65 @@ void ColorButton::init()
...
+			return qc;
+		else
+			return QColor(Qt::black); // default color is black.
+	}
+}
+
+bool ColorButton::isValidColor(const QColor& c)
+{
+	const QColor *ite = std::find(std::begin(colors), std::end(colors), c);
+	return (ite->isValid());
```
highperformancecoder: Dereferencing the end of a sequence (as what happens if the colour is not found) has undefined behaviour, and could even cause a null dereference crash.

The return test could be as simple as ite-color < colors_count, or to be more std libraryish ite!=std::end(colors)

libscidavis/src/ColorButton.h
```diff
@@ -35,6 +35,12 @@ class QPushButton;
 class QHBoxLayout;
 class QFrame;
 
+#if QT_VERSION >= 0x050000
+#define COLORNAME( c ) c.name(QColor::HexArgb)
+#else
+#define COLORNAME( c ) ('#'+QString::number(c.rgba(),16))
+#endif
+
```
highperformancecoder, line 43: inline function rather than macro here

libscidavis/src/ColorButton.h
```diff
@@ -52,6 +58,14 @@ class ColorButton : public QWidget
...
 	//! Get the color of the display part
 	QColor color() const;
+	//! Return the index for a given color
+	static unsigned int colorIndex(const QColor& c);
+	//! Return the color at index 'colorindex'
+	static QColor color(unsigned int colorIndex);
+	//! Returns TRUE if the color is included in the color box, otherwise returns FALSE.
+	static bool isValidColor(const QColor& color);
+	//! Returns the number of predefined colors
+	static int numPredefinedColors();
```
highperformancecoder: This is egregious encapsulation. colors_count is already a perfectly good constant public attribute.

libscidavis/src/PlotDialog.cpp
```diff
@@ -2388,14 +2336,13 @@ void PlotDialog::updateCanvasTransparency(int alpha)
...
-void PlotDialog::pickCanvasColor()
+void PlotDialog::pickCanvasColor(const QColor& c)
 {
-	QColor c = QColorDialog::getColor(boxCanvasColor->color(), this);
-    if ( !c.isValid() || c == boxCanvasColor->color() )
-			return;
-
-	boxCanvasColor->setColor ( c ) ;
-
+	QColor color = c;
```
highperformancecoder: Why not rename the argument color?

highperformancecoder: If color needs to be mutable, which is the likely reason, why not make the argument a QColor, and pass by value?

libscidavis/src/PlotDialog.cpp
```diff
@@ -2418,21 +2365,20 @@ void PlotDialog::pickCanvasColor()
...
+void PlotDialog::pickBackgroundColor(const QColor& c)
 {
-	QColor c = QColorDialog::getColor(boxBackgroundColor->color(), this);
-    if ( !c.isValid() || c == boxBackgroundColor->color() )
-			return;
-
-	boxBackgroundColor->setColor ( c ) ;
-
+	QColor color = c;
+	if (c.alpha() != boxBackgroundTransparency->value())
```
highperformancecoder, line 2377: ditto

highperformancecoder (changes requested):
As per code review comments, there a few niggles to pick.

libscidavis/src/ColorButton.cpp
gbm19: QColorDialog has 48 basic colors in a 6x8 grid. 
Using: 
 `for (int i=0; i < colors_count; i++) {QColorDialog::setStandardColor(i, colors[i].rgb());};`
arranges the colors in columns of 6. I  prefer to arrange them in rows of 8 colors each

libscidavis/src/ColorButton.cpp
gbm19: I have my doubts that there is someone able to distinguish that color #00000030 from white.
If someone complains, we will have to store colors with two variables (type and value).
Now, with very little effort we cover almost all #aarrggbb range.

gbm19: "... that color `#30` from white. "  
should be
"... that color (`#00000030`) from white."

libscidavis/src/ColorButton.cpp
gbm19: Qt takes care of all that trouble.

libscidavis/src/ColorButton.h
gbm19: Ok. See forthcoming commits.
My first intention was to require Qt5 to use only `QColor::HexArgb`. 
But I find it useful to have the color ehancement also for Qt4.

libscidavis/src/ColorButton.h
gbm19: Ok. I agree that `numPredefinedColors` can easily be removed.
I simply took needed code over from ColorBox.{h,cpp}.

libscidavis/src/ColorButton.cpp
highperformancecoder: Maybe a comment along the lines of rearranging the stand ard layout of standard colours might be useful. It puzzled me why you were only allocating half the colours.

The static_assert might also be useful, in case someone decides to remove some from the list, or monkey around with the magic numbers.

libscidavis/src/ColorButton.cpp
highperformancecoder: `#00000030` should be transparent actually. Mentally, I had the alpha channel trailing the colour channels, but with the # codes, its conventionally listed first.

However, the Qt documentation does not say how the 4 bytes are packed in an unsigned int. So does 0x30 mean `#00000030` (transparent) or `#30000000` (a very translucent shade of black)? It may depend on the machine endianess.

In principle, all shades of transparent are equivalent, and so that gives a wide range of codes to map to your standard colour table, but I'd think I'd shy away from mapping 0 (which is the obvious choice to represent transparent) to solid black, which is what you're currently doing.

libscidavis/src/ColorButton.cpp
highperformancecoder: I don't think so. std::find returns an iterator pointing to the end of the sequence, which in this case will be a pointer to one past the end the colour array. ite->isValid() is completely undefined - the compiler could initialise the memory past the end of the array to zero, or some garbage pattern, or it may put a properly initialised default QColor, or it may remain uninitialised to some random value.

libscidavis/src/ColorButton.h
highperformancecoder: I'm sure you've read Stroustrup on macros. Whilst he may egg the pudding a bit, it is widely accepted that macros should be avoided where alternatives exist. In this case an inline string function implementation of COLORNAME will do the job.

highperformancecoder: Ah OK - you fixed this.

libscidavis/src/ColorButton.cpp
```diff
@@ -70,11 +95,14 @@ ColorButton::ColorButton(QWidget *parent) : QWidget(parent)
 
 void ColorButton::init()
 {
+	// QColorDialog has 48 basic colors in a 6x8 grid. Using:
+	//   for (int i=0; i < colors_count; i++) {QColorDialog::setStandardColor(i, colors[i].rgb());};
+	// arranges the colors in columns of 6. Instead arrange them in rows of 8 colors each.
```
highperformancecoder, line 100: Your comment is still slightly confusing - given a 6x8 grid, columns of 6 is the same thing as rows of 8.
You could simply say "transpose the 6x8 grid", which is what you're actually doing.

libscidavis/src/ColorButton.cpp
highperformancecoder: I'm still keen on the static_assert:

static_assert(8*6 <= colours_count, "not enough default colors");

Better still, name the 8 and the 6, and use the names both in the loop and the assert. That way, as soon as anyone touches this in an invalid way, it will thow up a compiler error.

libscidavis/src/ColorButton.cpp
gbm19: I guess Qt appends a properly initialised QColor(Invalid) at end of array at the constructor, because that is what I get when I test the code.

libscidavis/src/ColorButton.cpp
gbm19: Qt docs  say explicitly "An ARGB quadruplet on the format #AARRGGBB, equivalent to an unsigned int." [http://doc.qt.io/archives/qt-4.8/qcolor.html#QRgb-typedef](http://doc.qt.io/archives/qt-4.8/qcolor.html#QRgb-typedef)

highperformancecoder: Yes - but it doesn't say how the individual bytes are packed into that unsigned int. My supposition is that it may depend on endianess - but I don't know...

highperformancecoder: I would expect that if the most and least significant byte were zero, then you have a guaranteed 2^16 different integer values mapping to transparent by stashing the values in the middle two bytes. I would avoid using 0, as that should definitely be reserved for transparent, but potentially the other 2^16-1 values are up for grabs...

libscidavis/src/ColorButton.cpp
highperformancecoder: It's not Qt. It's your compiler. Someone else's compiler may do it differently. Next year, your compiler may do it differently. It is not good to rely on it.

libscidavis/src/ColorButton.cpp
gbm19: But the use of inline COLORNAME allows us to ignore the details of using an uint to represent color. We save/load the color by its name, not by its uint.

highperformancecoder: I can't see the full code, so I'm missing your point. How is colorIndex to be used? At the moment, if I pass in QColor("#00000000") (ie transparent), colorIndex will return 0, (ie black). If colorIndex is private or protected, this may not matter much, of course, as perhaps there is no use case for passing transparent (or nearly transparent colours) to colorIndex. If it is a public method, then I think we do need to think how we deal with that possibility.

gbm19: QColor("#00000000") is as transparent as QColor("#00rrggbb") for any rgb value.
Thus, if you don't want to get 0 for colorIndex, use any value other than "#00000000"
Also, a transparent color is not very useful to represent data.
On the other hand, colorIndex() is only indispensable for the automatic color advance used when plotting several curves in the same graph.

gbm19:
Notice that last commit modifies the format of saved settings file (SciDAVis.conf)

highperformancecoder:
Re SciDAVis.conf, is the software backwardly compatible with the previous .conf file?

gbm19:
No. Code in this PR does not properly read  color values in previous version .conf file.
And code in [fc70ebc](https://github.com/highperformancecoder/scidavis/commit/fc70ebcf0a7b7511b762c44a09c44978b95d77ae) does not read .conf files written with this PR.
Until this PR there were 19 color values saved in SciDAVis.conf, 10 saved in "#rrggbb" format, and 9 saved in `QVariant` format. 
This PR stores all color values in "#aarrggbb" format.

Correction: 
This PR reads old .conf files correctly (thanks to QVariant.toString knowing it is dealing with color).
However, using the 1.23.x versions with new .conf files corrupts the settings file.

PR #106, opened 2018-10-23
Fix loading of color setttings in Qt4

gbm19:
Qt4 cannot read color values in `#aarrggbb` format. 
This PR solves the problem.

### DIFF
--- a/libscidavis/src/ColorButton.h
+++ b/libscidavis/src/ColorButton.h
@@ -37,8 +37,10 @@ class QFrame;
 
 #if QT_VERSION >= 0x050000
 static inline QString COLORNAME( QColor c ) { return c.name(QColor::HexArgb); }
+static inline QString COLORVALUE( QString s ) { return s; }
 #else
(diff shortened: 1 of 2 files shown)


libscidavis/src/ColorButton.h
```diff
@@ -37,8 +37,10 @@ class QFrame;
 
 #if QT_VERSION >= 0x050000
 static inline QString COLORNAME( QColor c ) { return c.name(QColor::HexArgb); }
+static inline QString COLORVALUE( QString s ) { return s; }
 #else
 static inline QString COLORNAME( QColor c ) { return ('#'+QString::number(c.rgba(),16)); }
+static inline QString COLORVALUE( QString s ) { if ((s[0]=='#') && (s.size()==9)) return s.remove(1,2); else return s; }
```
highperformancecoder, line 43: The downside of this approach is that all the colours are assumed to be opaque. But the workaround (extracting the alpha channel component and setting that separately) is probably not worth doing, given we're moving onto Qt5 shortly anyway,

PR #110, opened 2018-11-23
Two fixes and one enhancement

gbm19:


test/pythonTests/imageMarker.py
```diff
@@ -3,8 +3,7 @@
 image = layer.addImage("scidavis-logo.png")
 assert image.fileName()=="scidavis-logo.png"
 image.setSize(260,100)
-from PyQt4.QtCore import QSize 
-assert image.size()==QSize(260,100)
+assert image.size()==QtCore.QSize(260,100)
```
highperformancecoder, line 6: FYI, my thought is to deprecate all Qt/Qwt classes from the Python interface, as part of an effort to eliminate SIP/PyQt in favour of classdesc/boost::python, and ultimately get python support back in the windows build. 

In that case, ImageMarker.size() will most likely return a tuple. But that's for the future...

libscidavis/src/globals.cpp
```diff
@@ -159,7 +163,14 @@ void SciDAVis::about()
 	dialog->setAttribute(Qt::WA_DeleteOnClose);
 	dialog->setWindowTitle(QObject::tr("About SciDAVis"));
 	ui.version_label->setText(versionString() + extraVersion());
-	ui.release_date_label->setText(QObject::tr("Released") + ": " + QString(SciDAVis::release_date));
+	ui.release_date_label->setText(
+		QString("Qt: ")+QT_VERSION_STR+". Qwt: "+QWT_VERSION_STR
```
highperformancecoder, line 167: Excellent idea!

PR #113, opened 2018-12-02
Fix bug sf360. Plot associations button does nothing

gbm19:
Fix of sf #360 bug  for the Qt4 version

### DIFF
--- a/libscidavis/src/CurvesDialog.cpp
+++ b/libscidavis/src/CurvesDialog.cpp
@@ -212,10 +212,15 @@ void CurvesDialog::showPlotAssociations()
 		curve = 0;
 
     ApplicationWindow *app = (ApplicationWindow *)this->parent();
+#if QT_VERSION >= 0x050000
     close();
+#endif
 
     if (app)
         app->showPlotAssociations(curve);
+#if QT_VERSION < 0x050000
+    close();
+#endif
 }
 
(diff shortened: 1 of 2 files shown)


alxpl:
Confirmed, I just applied the patch in different versions of Fedora and the bug is fixed.

libscidavis/src/CurvesDialog.cpp
```diff
@@ -212,10 +212,15 @@ void CurvesDialog::showPlotAssociations()
 		curve = 0;
 
     ApplicationWindow *app = (ApplicationWindow *)this->parent();
```
highperformancecoder, line 214: Can we be guaranteed that the parent (if there is one) is an ApplicationWindow? Better to use a dynamic_cast here, I would think...

gbm19: In code there is only one use of the CurvesDialog constructor: In ApplicationWindow.cpp at line 5258:
`CurvesDialog* crvDialog = new CurvesDialog(this);`

highperformancecoder: That may be the case now, but there's nothing enforcing that assumption. For instance, someone may write a unittest for CurvesDialog...

I'm, merging the PR, but will change that line to a dynamic_cast. There's no performance reason for not using a dynamic_cast here.

PR #118, opened 2019-01-30
Fix bug sf #369: Fix random focus when deleting windows

gbm19:


libscidavis/src/PolynomialFit.cpp
```diff
@@ -174,29 +174,39 @@ void PolynomialFit::fit()
...
+		legend = "Y=" + QLocale().toString(d_results[0], 'g', d_prec);
+		for (unsigned j = 1; j < d_p; j++)
+		{
+			double cj = d_results[j];
+			if (cj>0 && !legend.isEmpty())
+				legend += "+";
+
+			QString s;
+			s.sprintf("%.5f",cj);
+			if (s != "1.00000")
```
highperformancecoder, line 197: This code is rather odd - why not use fabs(cj-1)>1e-5?

PR #120, opened 2019-03-13
Fix bug sf #377: All colors turn to black after reload of project

gbm19:


libscidavis/src/ColorButton.h
```diff
@@ -42,6 +42,7 @@ static inline QString COLORVALUE( QString s ) { return s; }
 static inline QString COLORNAME( QColor c ) { return ('#'+QString::number(c.rgba(),16)); }
 static inline QString COLORVALUE( QString s ) { if ((s[0]=='#') && (s.size()==9)) return s.remove(1,2); else return s; }
 #endif
+static inline uint    COLORUINT( QString s ) { if (s[0]=='#') return s.remove(0,1).toUInt(nullptr,16); else return s.toUInt(); }
```
highperformancecoder, line 45: With my HPC cap on, I would make the argument a const reference to be sure no copy of the string is made, as well as use strtoll(s.data()+1,nullptr,16) instead of the call to remove() to prevent an extra copy being created. Just a habit of mine.

Mind you, this is probably not performance critical code, so wouldn't matter anyway... :)

PR #125, opened 2019-08-18
Fix for bugs 383, 385.

gbm19:


libscidavis/src/ApplicationWindow.cpp
```diff
@@ -9647,19 +9648,23 @@ Graph* ApplicationWindow::openGraph(ApplicationWindow* app, MultiLayer *plot,
...
                       else
                         cl.penWidth = cl.lWidth;
-                      // custom dash pattern
-                      cl.lCapStyle = curve[16].toInt();
-                      cl.lJoinStyle = curve[17].toInt();
-                      cl.lCustomDash = curve[18];
+                      if (d_file_version >= 0x011800) // 1.24.0
+                        {
+                          // custom dash pattern
+                          cl.lCapStyle = curve[15].toInt();
```
highperformancecoder: This code bothers me. It would seem that curve[15] is penWidth, unless app->renamedTables is empty. ISTM, that you need a paramOffset variable, which is set to 15 when renamedTables is empty, and 16 otherwise, then calulate the following series of indices relative to paramOffset.

Of course, better still is to have a struct that names the relevant parameter fields, but that would be a refactor too far at this point in time, I suspect.

Unless, of course, renamedTables is guaranteed to be empty if d_file_version>=0x0118000, in which case an assert of that fact would be required.

libscidavis/src/ApplicationWindow.cpp
```diff
@@ -9647,19 +9648,23 @@ Graph* ApplicationWindow::openGraph(ApplicationWindow* app, MultiLayer *plot,
...
                       Table *w = app->table(curve[2]);
                       if (w)
                         {
                           int plotType = curve[3].toInt();
-                          if(curve.count()>21 && (plotType == Graph::VectXYXY || plotType == Graph::VectXYAM))
+                          if(curve.count()>(21+s_offset) && (plotType == Graph::VectXYXY || plotType == Graph::VectXYAM))
                             {
                               QStringList colsList;
-                              colsList<<curve[2]; colsList<<curve[20]; colsList<<curve[21];
+                              colsList<<curve[2]; colsList<<curve[20+s_offset]; colsList<<curve[21+s_offset];
```
highperformancecoder, line 9671: This solution is similar to what I was proposing for the previous problem...

libscidavis/src/RangeSelectorTool.cpp
```diff
@@ -129,11 +129,11 @@ void RangeSelectorTool::pointSelected(const QPoint &pos)
 			second_x = min_x;
 		if (second_x == max_x) { // start at selected point and try larger indices first
 			for (int i=0; i<n; ++i)
-				if (curve->x((i + point) % n) == max_x)
+				if (qFuzzyCompare(curve->x((i + point) % n), max_x))
```
highperformancecoder, line 132: I looked at the docs for qFuzzyCompare, and it didn't tell me what tolerance is used. Normally, for these sort of purposes, I roll my own functions along the lines of:
bool near(double x, double y, double tol) {
  double r=abs(x)+abs(y);
  if (r==0) return true;
  else return abs(x-y) < tol * r;
}
(PS - I haven't thought about NANs or Infs here)

Anyway, given that qFuzzyCompare must be practically indistinguishable from ==, I'll pass this use here.

highperformancecoder (changes requested):
Please fix up  offset of 15 or 16 depending on tables being empty or not.

libscidavis/src/ApplicationWindow.cpp
gbm19: ISTM that renamedTables does not change how curve parameters have to be read.

libscidavis/src/RangeSelectorTool.cpp
gbm19: You can get the definition of qFuzzyCompare in:
/usr/include/Qt/qglobal.h:2145 (for Qt 4.8.7)
/usr/include/qt5/QtCore/qglobal.h:827 (for Qt 5.12.4)

libscidavis/src/RangeSelectorTool.cpp
```diff
@@ -115,8 +115,8 @@ void RangeSelectorTool::pointSelected(const QPoint &pos)
 		setActivePoint(point);
 	else {
 		// try to be intelligent about selecting the inactive point
```
gbm19, line 117: Fix obtention of curve x-limits:
curve->minXValue(), curve->maxXValue() are obtained from curve->boundingRect() (see qwt_plot_curve.h in qwt source)
For curves with Symbols, curve->boundingRect() (see PlotCurve.cpp:345) is larger than curve->QwtPlotCurve::boundingRect() (see qwt_plot_curve.cpp:501)
Here we need to use the actual data limits, given by either curve->data().boundingRect() or curve->QwtPlotCurve::boundingRect()

libscidavis/src/ApplicationWindow.cpp
highperformancecoder: Looks good!

libscidavis/src/RangeSelectorTool.cpp
highperformancecoder: Very similar to my near function, with a tolerance of 1e-12.

highperformancecoder (approved):

PR #132, opened 2020-01-29
Suggested fix for different location of .sip

srtlg:
While trying to pack scidavis-qt5 for Arch Linux, I came across a problem with
the hard-coded location of the .sip files in python-sipcmd.py. I don't know whether
the suggested fix works for distributions other than Arch (Python 3.8.1, PyQt5 version 5.14.1)

python2-pyqt5
/usr/share/sip/PyQt5/QtCore/QtCoremod.sip
python-pyqt5
/usr/lib/python3.8/site-packages/PyQt5/bindings/QtCore/QtCoremod.sip

Furthermore, it is really useful if qmake just gives up when sip fails

### DIFF
--- a/libscidavis/python-sipcmd.py
+++ b/libscidavis/python-sipcmd.py
@@ -26,6 +26,8 @@
 ############################################################################
 
 
+import os
+import warnings
 import sys, sipconfig
 config = sipconfig.Configuration()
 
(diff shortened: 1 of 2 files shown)


highperformancecoder:
Thank you for your contribution. However, the plan is to ditch PyQt and sip altogether for SciDaVis2. I am really hoping not to have to do another SciDaVis 1.x release :)

alxpl:
I think the check
`if sys.version_info[0] >= 3 and sys.version_info[1] >= 7 and pyqt == 'PyQt5':`
should be changed to either
`if sys.version_info[0] >= 3 and sys.version_info[1] >= 8 and pyqt == 'PyQt5':`
or
`if sys.version_info[0] >= 3 and sys.version_info[1] > 7 and pyqt == 'PyQt5':`

I did some builds of scidavis and with the check as it is, only fedora versions with Python >= 3.8 succeed. I looked a bit into it and the `importlib.metadata` thingy was added in 3.8. We do have a module for 3.7 that offers the same functionality, but I think the syntax may be different, because I could not get it to work.

alxpl:
Just confirmed it with our resident Python expert, importlib.metadata is 3.8+.

PR #133, opened 2020-02-15
Update python-sipcmd.py

alxpl:
importlib.metadata will only work with python 3.8 and newer.


highperformancecoder:
Fine. Thanks for the PR. We're doing away with SIP for SciDAVis 2.x, though, so this may never appear in a release.

alxpl:
> Fine. Thanks for the PR. We're doing away with SIP for SciDAVis 2.x, though, so this may never appear in a release.

I know, but I just could not leave it there, it's probably my OCD :smiley:

Issue #130, opened 2019-09-27
Building under Arch (GCC 9.1.0, Qt 5.13.1) fails with `reference to 'future' is ambiguous'


srtlg:
There seems to be a conflict between your future namespace and std::future of C++11

[This patch](https://aur.archlinux.org/cgit/aur.git/tree/future_absolute.patch?h=scidavis-qt5) solves the issue and is against [version 1.25.1](https://github.com/highperformancecoder/scidavis/commit/0f3383dc57bc9b756a65d90e88da91952eeb6173)

```
In file included from src/ApplicationWindow.cpp:102:
src/future/core/Project.h:46:24: error: reference to ‘future’ is ambiguous
   46 | class Project : public future::Folder
      |                        ^~~~~~
In file included from /usr/include/qt/QtCore/qthread.h:50,
                 from /usr/include/qt/QtCore/QtCore:234,
                 from /usr/include/qt/QtOpenGL/QtOpenGLDepends:3,
                 from /usr/include/qt/QtOpenGL/QtOpenGL:3,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_openglhelper.h:5,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_types.h:26,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_drawable.h:7,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_label.h:10,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_axis.h:5,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_coordsys.h:4,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_plot.h:4,
                 from /usr/include/qt/qwtplot3d-qt5/qwt3d_surfaceplot.h:4,
                 from src/Graph3D.h:32,
                 from src/ApplicationWindow.cpp:59:
/usr/include/c++/9.1.0/future:125:11: note: candidates are: ‘template<class _Res> class std::future’
  125 |     class future;
      |           ^~~~~~
```

highperformancecoder:
That patch looks fine - is there any chance you could do this as a pull request?

srtlg:
There are a lot more references to the future namespace, do you think, that all need to be made absolute?
(It seems, that QThread is the culprit in my case, it includes future)

PR #134, opened 2020-03-06
compiles under arch

srtlg:
Hello, I've tried to package the classdesc-python branch for Arch Linux. The following changes seem to me reasonable for including into upstream.

- There are again conflicts with references to `std::future` like #130 
- in `scidavis/scidavis.pro` I do not get `-lpython...` without `python-config --libs --embed`

### DIFF
--- a/libscidavis/src/ApplicationWindow.cpp
+++ b/libscidavis/src/ApplicationWindow.cpp
@@ -14042,7 +14042,7 @@ void ApplicationWindow::showHistory()
 
 QStringList ApplicationWindow::tableWindows()
 {
-	QList<AbstractAspect *> tables = d_project->descendantsThatInherit("::future::Table");
+	QList<AbstractAspect *> tables = d_project->descendantsThatInherit("future::Table");
 	QStringList result;
 	foreach(AbstractAspect *aspect, tables)
           result.append(aspect->name().c_str());
--- a/libscidavis/src/ApplicationWindow.cpp
+++ b/libscidavis/src/ApplicationWindow.cpp
@@ -2655,7 +2655,7 @@ TableStatistics *ApplicationWindow::newTableStatistics(Table *base, int type, QL
 
 void ApplicationWindow::removeDependentTableStatistics(const AbstractAspect *aspect)
 {
-	future::Table *future_table = qobject_cast<future::Table *>(const_cast<AbstractAspect *>(aspect));
+	::future::Table *future_table = qobject_cast<::future::Table *>(const_cast<AbstractAspect *>(aspect));
 	if (!future_table) return;
 	QList<MyWidget*> windows = windowsList();
 	foreach(MyWidget *win, windows)
@@ -13984,13 +13984,13 @@ void ApplicationWindow::selectPlotType(int type)
 void ApplicationWindow::handleAspectAdded(const AbstractAspect *parent, int index)
 {
 	AbstractAspect * aspect = parent->child(index);
-	future::Matrix * matrix = qobject_cast<future::Matrix *>(aspect);
+	::future::Matrix * matrix = qobject_cast<::future::Matrix *>(aspect);
 	if (matrix)
 	{
 		initMatrix(static_cast<Matrix *>(matrix->view()));
 		return;
 	}
-	future::Table * table = qobject_cast<future::Table *>(aspect);
+	::future::Table * table = qobject_cast<::future::Table *>(aspect);
 	if (table)
 	{
 		initTable(static_cast<Table *>(table->view()));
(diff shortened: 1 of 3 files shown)


libscidavis/src/ApplicationWindow.cpp
```diff
@@ -14042,7 +14042,7 @@ void ApplicationWindow::showHistory()
 
 QStringList ApplicationWindow::tableWindows()
 {
-	QList<AbstractAspect *> tables = d_project->descendantsThatInherit("future::Table");
+	QList<AbstractAspect *> tables = d_project->descendantsThatInherit("::future::Table");
```
highperformancecoder: Is this change necessary?

scidavis/scidavis.pro
```diff
@@ -17,7 +17,7 @@ python {
   }
   unix: {
         LIBS+=$$system($$PYTHONBIN findBoostPythonLib.py)
-        LIBS+=$$system($$PYTHONBIN-config --ldflags)
+        LIBS+=$$system($$PYTHONBIN-config --libs --embed)
```
highperformancecoder: --embed doesn't work on my system. It causes an error. And --ldflags also includes the flags -Xlinker -export-dynamic. I can't find documentation on what is supposed to be the difference between --libs and --ldflags - do you know?

srtlg:
On Fri, Mar 06, 2020 at 14:19 -0800, Russell Standish wrote:
> > @@ -17,7 +17,7 @@ python {
>    }
>    unix: {
>          LIBS+=$$system($$PYTHONBIN findBoostPythonLib.py)
> -        LIBS+=$$system($$PYTHONBIN-config --ldflags)
> +        LIBS+=$$system($$PYTHONBIN-config --libs --embed)
>
> --embed doesn't work on my system. It causes an error. And --ldflags also includes the flags -Xlinker -export-dynamic. I can't find documentation on what is supposed to be the difference between --libs and --ldflags - do you know?
Sorry, I don't know much about python-config. I just deduced from the error messages (linking scidavis produced lots of unresolved symbols that were all in pythonX.Y.so)

$ python-config --ldflags
 -L/usr/lib  -lcrypt -lpthread -ldl  -lutil -lm -lm
$ python-config --libs
 -lcrypt -lpthread -ldl  -lutil -lm -lm
$ python-config --embed
$ python-config --libs --embed
-lpython3.8 -lcrypt -lpthread -ldl  -lutil -lm -lm

srtlg:
I've just checked. That is a Python 3.8 thing.

https://docs.python.org/3/whatsnew/3.8.html

"To embed Python into an application, a new --embed option must be passed to
python3-config --libs --embed to get -lpython3.8 (link the application to
libpython). To support both 3.8 and older, try python3-config --libs --embed
first and fallback to python3-config --libs (without --embed) if the previous
command fails."

They altered Python*/Misc/python-config.sh.in between 3.7 and 3.8 to have this effect
From 3.8 onwards only --embed will add -lpythonX.Y to the flags

scidavis/scidavis.pro
srtlg: What about this patch? It works under Fedora 30 (Python 3.7) and Arch (Python 3.8)

```patch
+diff -u10 -r scidavis-2.0.10-original/scidavis/scidavis.pro scidavis-2.0.10/scidavis/scidavis.pro
+--- scidavis-2.0.10-original/scidavis/scidavis.pro     2020-02-17 07:54:00.000000000 +0100
++++ scidavis-2.0.10/scidavis/scidavis.pro      2020-03-07 11:27:31.630929748 +0100
+@@ -10,21 +10,25 @@

  include(../config.pri)
  include( basic.pri )
  python {
    PYTHONBIN = $$(PYTHON)
    isEmpty( PYTHONBIN ) {
      PYTHONBIN = python3
    }
    unix: {
          LIBS+=$$system($$PYTHONBIN findBoostPythonLib.py)
 -        LIBS+=$$system($$PYTHONBIN-config --ldflags)
-+        LIBS+=$$system($$PYTHONBIN-config --libs --embed)
++        system($$PYTHONBIN-config --embed >/dev/null) {
++            LIBS+=$$system($$PYTHONBIN-config --libs --embed)
++        } else {
++            LIBS+=$$system($$PYTHONBIN-config --libs)
++        }
        }
    message($$LIBS)

    compiledScidavisrc.target = scidavisrc.pyc
    compiledScidavisrc.depends = scidavisrc.py
    compiledScidavisrc.commands = $$PYTHONBIN -m py_compile scidavisrc.py
    compiledScidavisUtil.target = scidavisUtil.pyc
    compiledScidavisUtil.depends = scidavisUtil.py
    compiledScidavisUtil.commands = $$PYTHONBIN -m py_compile scidavisUtil.py
    QMAKE_EXTRA_TARGETS += compiledScidavisrc compiledScidavisUtil
```

libscidavis/src/ApplicationWindow.cpp
srtlg: you are right. It has to remain `descendantsThatInherit("future::Table");` 
I tested it the following way
```
d_project->descendantsThatInherit("::future::Table").length() is always zero
d_project->descendantsThatInherit("future::Table").length() gives the correct number of Tables
```

scidavis/scidavis.pro
highperformancecoder: That looks like it will work. Looks like there was a change here between Python 3.7 and 3.8.

libscidavis/src/ApplicationWindow.cpp
highperformancecoder: If you remove the change on this line, we're good to go on this PR.

PR #140, opened 2020-06-20
Add CMake build support.

ghost:
Missing:
- Self contained
- Python
- Test


highperformancecoder:
Thanks for this. Personally, I'd prefer using cmake to qmake, and these look like good clean config files.

highperformancecoder:
I'll try to bring the rest of the dev team over to cmake...

PR #141, opened 2020-06-21
Fix Compilation with MSVC 2019.

ghost:


libscidavis/src/Convolution.cpp
```diff
@@ -138,7 +138,7 @@ void Convolution::addResultCurve()
 
   d_table->addCol();
   d_table->addCol();
-  double x_temp[d_n];
+  QVarLengthArray<double> x_temp(d_n);
```
highperformancecoder: Why not use std::vector?

libscidavis/src/Spectrogram.h
```diff
@@ -99,9 +99,12 @@ class MatrixData: public QwtRasterData
 	d_m = new double* [n_rows];
 	for ( int l = 0; l < n_rows; ++l)
 		d_m[l] = new double [n_cols];
-
-        min_z=std::numeric_limits<double>::max();
-        max_z=-std::numeric_limits<double>::max();
+#ifdef _MSC_VER
+        min_z = (std::numeric_limits<double>::max)();
```
highperformancecoder: Why are parentheses needed?

libscidavis/src/Convolution.cpp
highperformancecoder: QVarLengthArray has certain performance advantages over std::vector for small values of d_n, and also doesn't initialise the data. OK the latter is probably worthwhile, even if the data still ends up being on the heap.

libscidavis/src/Spectrogram.h
ghost: MSVC STL library `min/max` implementation doesn't conform to standard STL.
The only way to use them is to put them between parentheses.
see https://github.com/microsoft/STL/blob/master/stl/inc/limits

highperformancecoder:
On Sat, Jun 27, 2020 at 12:03:07PM -0700, مهدي شينون (Mehdi Chinoune) wrote:
> @ChinouneMehdi commented on this pull request.
> 
> 
> In libscidavis/src/Spectrogram.h:
> 
> > @@ -99,9 +99,12 @@ class MatrixData: public QwtRasterData
>         d_m = new double* [n_rows];
>         for ( int l = 0; l < n_rows; ++l)
>                 d_m[l] = new double [n_cols];
> -
> -        min_z=std::numeric_limits<double>::max();
> -        max_z=-std::numeric_limits<double>::max();
> +#ifdef _MSC_VER
> +        min_z = (std::numeric_limits<double>::max)();
> 
> MSVC STL library min/max implementation doesn't conform to standard STL.
> The only way to use them is to put them between parentheses.
> see https://github.com/microsoft/STL/blob/master/stl/inc/limits

I still don't understand why. What happens if you don't put them in parentheses?

Also:

    _NODISCARD static constexpr double(min)() noexcept {
        return DBL_MIN;
    }

doesn't look syntactically correct for C++. But it compiles on g++,
and seems to be equivalent to what you get when leaving the
parentheses off around min. I'm guessing it declares a variable (or in
this case a member) min that has a function type (thus is callable). I
only say this, as the declaration "double (*min)()" decalres a
function pointer.


-- 

                      http://www.hpcoders.com.au

libscidavis/src/Spectrogram.h
highperformancecoder: Actually, the better link is [this answer on StackOverflow](https://stackoverflow.com/questions/40492414/why-does-stdnumeric-limitslong-longmax-fail).  It is caused by VC++ using min/max macros. There are two solutions - put the parentheses around the min/max, which appears to be acceptable in g++ as well, or define NOMINMAX immediately after the #include <limits> statement. My preference is the latter, as its less confusing, and more standards conformant, but if the former, we don't need the #ifdef MSC_VER guard,

ghost: done.

PR #143, opened 2020-06-30
Silent warnings from different compilers.

ghost:
Compilers:
 - MSVC 2019
 - Clang 10.0.0
 - GCC 9.3.0/10.1.0
with Qt 5.12.8

### DIFF
--- a/libscidavis/src/FFT.cpp
+++ b/libscidavis/src/FFT.cpp
@@ -121,7 +121,7 @@ QList<Column *> FFT::fftTable()
 
 	for(size_t i=0; i<d_n; i++)
 	{
-		int i2 = 2*i;
+        size_t i2 = 2*i;
 		double a = sqrt(d_y[i2]*d_y[i2] + d_y[i2+1]*d_y[i2+1]);
 		amp[i]= a;
 		if (a > aMax)
@@ -132,7 +132,7 @@ QList<Column *> FFT::fftTable()
 	columns << new Column(tr("Imaginary"), SciDAVis::Numeric);
 	columns << new Column(tr("Amplitude"), SciDAVis::Numeric);
 	columns << new Column(tr("Angle"), SciDAVis::Numeric);
-	for (size_t i=0; i<d_n; i++)
+    for (int i=0; i<static_cast<int>(d_n); i++)
 	{
 		int i2 = 2*i;
 		columns.at(0)->setValueAt(i, d_x[i]);
--- a/libscidavis/src/Graph.cpp
+++ b/libscidavis/src/Graph.cpp
@@ -1564,7 +1564,7 @@ void Graph::removeMarker()
 		if (selectedMarker==legendMarkerID)
 			legendMarkerID=-1;
 
-		if (d_lines.contains(selectedMarker)>0)
+        if (d_lines.contains(selectedMarker))
 		{
 			int index = d_lines.indexOf(selectedMarker);
             int last_line_marker = (int)d_lines.size() - 1;
@@ -1572,7 +1572,7 @@ void Graph::removeMarker()
 				d_lines[i]=d_lines[i+1];
 			d_lines.resize(last_line_marker);
 		}
-		else if(d_texts.contains(selectedMarker)>0)
+        else if(d_texts.contains(selectedMarker))
 		{
 			int index=d_texts.indexOf(selectedMarker);
 			int last_text_marker = d_texts.size() - 1;
@@ -1580,7 +1580,7 @@ void Graph::removeMarker()
 				d_texts[i]=d_texts[i+1];
 			d_texts.resize(last_text_marker);
 		}
-		else if(d_images.contains(selectedMarker)>0)
+        else if(d_images.contains(selectedMarker))
 		{
 			int index=d_images.indexOf(selectedMarker);
 			int last_image_marker = d_images.size() - 1;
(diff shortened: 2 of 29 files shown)


libscidavis/src/ApplicationWindow.h
```diff
@@ -179,7 +179,9 @@ public slots:
   ///* load project file \a into this
   ///* @return true if project load successful
   bool loadProject(const QString& fn);
+#ifdef ORIGIN_IMPORT
   ApplicationWindow* importOPJ(const QString& filename);
+#endif
```
highperformancecoder: This sort of change causes problems with classdesc, and will be a problem in the SciDaVis 3 build.

libscidavis/src/FFT.cpp
```diff
@@ -117,11 +117,11 @@ QList<Column *> FFT::fftTable()
 	{
 		for(size_t i=0; i<d_n; i++)
 			d_x[i] = i*df;
-	}
+    }
 
-	for(size_t i=0; i<d_n; i++)
-	{
-		int i2 = 2*i;
+    for(int i=0; i < static_cast<int>(d_n); i++)
```
highperformancecoder: Why make the index variable signed?

libscidavis/src/Graph.cpp
```diff
@@ -2225,7 +2225,7 @@ QString Graph::saveTitle()
 
 QString Graph::saveScaleTitles()
 {
-	int a;
+    int a{};
```
highperformancecoder, line 2228: Probably a=0 is clearer, but equivalent.

libscidavis/src/MuParserScript.cpp
```diff
@@ -155,7 +155,7 @@ MuParserScript::MuParserScript(ScriptingEnv *environment, const QString &code, Q
 
 	// statement separation needs lower precedence than everything else; assignment has precedence
 	// -1, everything else defined in mu::Parser has non-negative precedence
-	m_parser.DefineOprt(_T(";"), statementSeparator, -2);
+    m_parser.DefineOprt(_T(";"), statementSeparator, 0);
```
highperformancecoder, line 156: I assume this is a change to muparser, and doesn't break anything. But the comment above now doesn't make sense.

libscidavis/src/PlotCurve.cpp
```diff
@@ -155,59 +155,7 @@ QList< QVector<double> > DataCurve::convertData(const QList<Column*> &cols, cons
...
-							if (time.isValid()) break;
-						}
-					}
-
-					if (format.isEmpty())
-						format = static_cast<DateTime2StringFilter *>(col->outputFilter())->format();
-
-					reference_dates.push_back(QDate());
-					reference_times.push_back(time);
-					if (g)
```
highperformancecoder, line 179: I'm glad to see someone is brave enough to make this change. I've been too lily-livered myself :).

libscidavis/src/future/lib/Interval.h
```diff
@@ -42,13 +42,18 @@ template<class T> class IntervalBase
...
 			d_start = other.start();
 			d_end = other.end();
 		}
-		IntervalBase(T start, T end) {
-			d_start = start;
-			d_end = end;
+		IntervalBase<T>& operator=(const IntervalBase<T>& other) {
+			d_start = other.start();
+			d_end = other.end();
+			return *this;
```
highperformancecoder: We should use the default assignment and copy constructor here.

libscidavis/src/FFT.cpp
ghost: MSVC warns about `i2 = 2*i` because of mismatch size_t/int.

libscidavis/src/ApplicationWindow.h
ghost: Could you elaborate more.

libscidavis/src/MuParserScript.cpp
ghost: Done.

libscidavis/src/ApplicationWindow.h
highperformancecoder: Classdesc processes C++ code, but can't handle the preprocessor. Classdesc doesn't know what the value of ORIGIN_IMPORT is. In this case, it will generate a reference to ApplicationWindow::importOPJ, even when ORIGIN_IMPORT is undefined, causing a compile error.

Best is to avoid #if constructions in header files.

libscidavis/src/FFT.cpp
highperformancecoder: Couldn't we make i2 unsigned? I can't see what the rest of the code is doing in this code review.

libscidavis/src/ApplicationWindow.h
ghost: Maybe Classdesc should be fixed.

highperformancecoder: Its a tradeoff. Classdesc cannot know the values of all the macros seen by the compiler, and if you preprocess the source code, you end up with classdesc generating a huge amount of unnecessary code that a) blows compile times by an order of magnitude, and b) adds a large amount of extra handling by the programmer, which is not necessary. All just to have conditional compilation in a header. Not worth the tradeoff.

Maybe if/when C++ gets first class reflection support in the language, then this restriction will be unnecessary. Reflection didn't make the cut for C++20 :(.

PR #144, opened 2020-07-01
Replace/Remove declarations deprecated by Qt 5.15

ghost:


ghost:
Ping!

highperformancecoder:
I started trying to merge this yesterday. I had fix a number of compilation errors in the Windows build. But more importantly, scidavis crashes when trying to plot a simple plot. Not sure if this is in your code, or in an updated dependent library. I haven't built scidavis in a while.

ghost:
> 
> 
> I started trying to merge this yesterday. I had fix a number of compilation errors in the Windows build. But more importantly, scidavis crashes when trying to plot a simple plot. Not sure if this is in your code, or in an updated dependent library. I haven't built scidavis in a while.

I can't reproduce the crash on Windows with both MinGW-w64 and MSVC.

libscidavis/src/AssociationsDialog.cpp
```diff
@@ -131,7 +131,11 @@ void AssociationsDialog::changePlotAssociation(int curve, const QString& text)
   if (c->plotAssociation() == text)
     return;
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+  QStringList lst = text.split(",", Qt::SkipEmptyParts);
+#else
   QStringList lst = text.split(",", QString::SkipEmptyParts);
+#endif
```
gbm19, line 138: Wouldn’t it be better (less intrusive) to use macros?
For instance:
```suggestion
#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
#define QT_SKIPEMPTYPARTS Qt::SkipEmptyParts
#define QT_KEEPEMPTYPARTS Qt::KeepEmptyParts
#else
#define QT_SKIPEMPTYPARTS QString::SkipEmptyParts
#define QT_KEEPEMPTYPARTS QString::KeepEmptyParts
#endif
```

ghost: > Wouldn’t it be better (less intrusive) to use macros?
> For instance:
> ```suggestion
> #if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
> #define QT_SKIPEMPTYPARTS Qt::SkipEmptyParts
> #define QT_KEEPEMPTYPARTS Qt::KeepEmptyParts
> #else
> #define QT_SKIPEMPTYPARTS QString::SkipEmptyParts
> #define QT_KEEPEMPTYPARTS QString::KeepEmptyParts
> #endif
> ```

And where to put these macros?

gbm19: Either in a common header or in the '*.cpp' files that use the macros. 
That is what I did for the python2/python3 switch in files PythonScript.cpp, PythonScripting.cpp, scidavis.sip

highperformancecoder:
The crash was occurring in a Linux build (OpenSUSE Tumbleweed). I
haven't tested it on Windows, just the build, which is done using the
MXE cross compiler on Linux.

On Tue, Jul 07, 2020 at 01:53:28AM -0700, مهدي شينون (Mehdi Chinoune) wrote:
>     I started trying to merge this yesterday. I had fix a number of compilation
>     errors in the Windows build. But more importantly, scidavis crashes when
>     trying to plot a simple plot. Not sure if this is in your code, or in an
>     updated dependent library. I haven't built scidavis in a while.
> 
> I can't reproduce the crash on Windows with both MinGW-w64 and MSVC.
> 
> —
> You are receiving this because you commented.
> 

-- 

                      http://www.hpcoders.com.au

highperformancecoder:
The crash occurs on Tumbleweed. It compiles and runs fine on Ubuntu Bionic. I think the Tumbleweed problem may be due to a faulty Qwt or QwtPlot build, so will need to dig into that.

PR #145, opened 2020-07-07
Adapt to use system libraries. Use standard variable names.

gbm19:
This PR allows me to compile SciDAVis using the system libs in a Fedora 30  distribution with qwt5-qt5-devel, qwtplot3d-qt5-devel, liborigin-devel and (unused) qwt5-qt4-devel, qwtplot3d-qt4-devel, qwt-qt5-devel-6.1.3 packages installed simultaneously. No python yet.


3rdparty/CMakeLists.txt
```diff
@@ -7,16 +7,31 @@ set( CMAKE_AUTORCC OFF )
 add_library( minigzip "minigzip/minigzip.c" )
 target_link_libraries( minigzip ZLIB::ZLIB )
 
+if (Qt5Gui_FOUND)
+  get_target_property(QT_INCLUDE_DIR Qt5::Gui INTERFACE_INCLUDE_DIRECTORIES)
+  message( STATUS "Qt5 GUI found ${QT_INCLUDE_DIR}" )
+else()
+  message( STATUS Qt5 GUI not found )
+endif ()
```
ghost, line 15: Qt::Gui was declared as a required component, so cmake will throw an error message if Qt::Gui wasn't found.
There is no need for this.

gbm19: It doesn't work without it.
You need at least the second line (get_target_property) to fill in the QT_INCLUDE_DIR variable to use it 
for the HINTS in 'findpath' command for QWT_INCLUDE_DIR and QWTPLOT3D_INCLUDE_DIR.
The third line (message) shows you where qwt_plot.h and qwt3d_plot.h will be looked for.

ghost: > 
> It doesn't work without it.
> You need at least the second line (get_target_property) to fill in the QT_INCLUDE_DIR variable to use it 
> for the HINTS in 'findpath' command for QWT_INCLUDE_DIR and QWTPLOT3D_INCLUDE_DIR.
> The third line (message) shows you where qwt_plot.h and qwt3d_plot.h will be looked for.
> 

There is no need to check if Qt5::Gui is found because cmake will abort immediately if Qt5::Gui wasn't found.

highperformancecoder: I think I understand Miquel's argument here, so I'm allowing this PR through. The proof will be when I do the OBS integration...

PR #146, opened 2020-07-12
Add python support to cmake build

gbm19:


CMakeLists.txt
```diff
@@ -63,6 +63,8 @@ endif()
 
 option( SCRIPTING_MUPARSER "Enable muParser Scripting" ON )
 
+option( ENABLE_python "Enable Python binding" ON )
```
ghost: Isn't ENABLE_PYTHON better?
Binding or Scripting?

cmake/scidavis-python.cmake
```diff
@@ -0,0 +1,179 @@
...
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor,
+#  Boston, MA  02110-1301  USA
+
+# based on:
+#  https://github.com/PLplot/PLplot/blob/master/cmake/modules/qt.cmake
+#  https://github.com/PLplot/PLplot/blob/master/bindings/qt_gui/pyqt5/CMakeLists.txt
+#  http://cmake.3232098.n2.nabble.com/cmake-PyQT-SIP-td5789749.html
+
+if( ENABLE_python )
+  find_package(PythonInterp 3)
```
ghost: According to cmake [documentation](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html?highlight=findpython#module:FindPythonInterp) this is a deprecated module, you should use FindPython3 instead.

libscidavis/CMakeLists.txt
```diff
@@ -404,6 +419,14 @@ if( ORIGIN_IMPORT )
   endif()
 endif()
 
+if( ENABLE_python )
+  target_compile_definitions( libscidavis PUBLIC
+    PYTHON_CONFIG_PATH="/etc"
+    PYTHON_UTIL_PATH="/usr/share/scidavis"
+    SCRIPTING_PYTHON
+    )
```
ghost, line 427: How about Windows?

cmake/scidavis-python.cmake
```diff
@@ -0,0 +1,179 @@
...
+
+  set(scidavis_pyqt5_HDR ${CMAKE_CURRENT_BINARY_DIR}/sipAPIscidavis.h)
+
+  set_source_files_properties(
+    ${scidavis_pyqt5_HDR}
+    PROPERTIES
+    GENERATED ON
+    )
+
+  set(scidavis_pyqt5_SRC
```
ghost, line 101: SCIDAVIS_PYQT5_SRCS

ghost (changes requested):

cmake/scidavis-python.cmake
highperformancecoder: Makes sense

cmake/scidavis-python.cmake
highperformancecoder: Could you justify why CMake variables should all be upper case? Seems like it is just a style thing, and often the choice is opposite (free-case).

libscidavis/CMakeLists.txt
highperformancecoder: Currenly Python is not supported on Windows because MXE (a Windows cross compiler running on Linux) doesn't support it. I'm working on that (PyQt is probably too large a bridge to cross with MXE, hence the Classdesc-based reflection approach of SciDaVis 3.x).

Nevertheless, if you want to get Python to work on Visual Studio builds, knock yourself out!

cmake/scidavis-python.cmake
ghost: I don't have a strong argument, just a style coding convention that I saw in many projects.

libscidavis/CMakeLists.txt
ghost: So python support should be disabled on Windows.

CMakeLists.txt
gbm19: Actually both. Binding is the hard one to get it properly.

cmake/scidavis-python.cmake
gbm19: See next commit

cmake/scidavis-python.cmake
gbm19: I just adapted the naming of PLplot.

libscidavis/CMakeLists.txt
highperformancecoder: Only on the MXE target. If you can get it to work on Visual Studio, you're welcome to enable it, but I for one, won't be supporting that configuration.

cmake/scidavis-python.cmake
highperformancecoder: OK - let's leave it as is for now. I'm not a fan of overly prescriptive style guides.

alxpl:
Will cmake replace make or will they both be available?

ghost:
> Will cmake replace make or will they both be available?

You mean qmake.
Because cmake is just a generator. You can generate Makefile, Ninja-build, Visual studio project, Xcode project, ...etc.

alxpl:
On Wed, Jul 15, 2020, 11:12 مهدي شينون (Mehdi Chinoune) <
notifications@github.com> wrote:

> Will cmake replace make or will they both be available?
>
> You mean qmake.
> Because cmake is just a generator. You can generate Makefile, Ninja-build,
> Visual studio project, Xcode project, ...etc.
>

Yes, sorry, I wanted to write qmake, but at the moment I had just finished
working on a spec file with %make macros.
The thing is that as package maintainers (Fedora here) we are supposed to
stay close to upstream, so if everything is going to be built and tested
using cmake, I should switch as well.

highperformancecoder:
On Wed, Jul 15, 2020 at 02:27:40AM -0700, alxpl wrote:
> On Wed, Jul 15, 2020, 11:12 مهدي شينون (Mehdi Chinoune) <
> notifications@github.com> wrote:
> 
> > Will cmake replace make or will they both be available?
> >
> > You mean qmake.
> > Because cmake is just a generator. You can generate Makefile, Ninja-build,
> > Visual studio project, Xcode project, ...etc.
> >
> 
> Yes, sorry, I wanted to write qmake, but at the moment I had just finished
> working on a spec file with %make macros.
> The thing is that as package maintainers (Fedora here) we are supposed to
> stay close to upstream, so if everything is going to be built and tested
> using cmake, I should switch as well.
> 

That's the plan. There's still a bit to go porting over the relevant
options from qmake to cmake.


-- 

                      http://www.hpcoders.com.au

PR #153, opened 2020-08-31
Four bug fixes, including sf #402 and #403

gbm19:


libscidavis/src/Graph.cpp
```diff
@@ -5332,6 +5332,10 @@ void Graph::updateCurveNames(const QString& oldName, const QString& newName, boo
         if (it->rtti() != QwtPlotItem::Rtti_PlotCurve)
             continue;
 
+        PlotCurve *p = (PlotCurve *)it;
```
highperformancecoder: Line 5332 guarantees this to work, I believe, but it would be clearer to use dynamic_cast instead. It's more standard.

libscidavis/src/MuParserScript.cpp
```diff
@@ -140,21 +140,15 @@ MuParserScript::MuParserScript(ScriptingEnv *environment, const QString &code, Q
...
 	m_parser.SetVarFactory(variableFactory, this);
 
-	// redefine characters for operators to include ";"
         static const auto opChars=
           // standard operator chars as defined in mu::Parser::InitCharSets()
           _T("abcdefghijklmnopqrstuvwxyz")
           "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
-          "+-*^/?<>=#!$%&|~'_"
-          // our additions
-          ";";
```
highperformancecoder, line 150: What's the impact of removing the ';' ?

highperformancecoder:
Could you please point me in the direction of the bug report that removing the ';' operator from muparser solves?
Also could you at least mark the bugs fixed as 'fixed' or even 'closed' if you're really confident that they've been addressed. Otherwise, we'll be in a situation in a year or two's time having to test all the tickets to see whether they've been resolved.

libscidavis/src/MuParserScript.cpp
gbm19: Almost none. What was done with ";" can be done with ",".
The signature of mu::Parser::DefineOprt is:
`/usr/include/muParserBase.h`
`138:    void DefineOprt(const string_type &a_strName, `
`139-                    fun_type2 a_pFun, `
`140-                    unsigned a_iPri=0, `
`141-                    EOprtAssociativity a_eAssociativity = oaLEFT,`
`142-                    bool a_bAllowOpt = false);`

Commit 62d4e5f (after 2.3.0) contains this code:
`--- a/libscidavis/src/MuParserScript.cpp`
`+++ b/libscidavis/src/MuParserScript.cpp`
`@@ -153,9 +153,7 @@ MuParserScript::MuParserScript(ScriptingEnv *environment, const QString &code, Q`
`         // work around muparser bug number 6 https://code.google.com/p/muparser/issues/detail?id=6`
` 	m_parser.DefineInfixOprtChars(opChars);`
`-	// statement separation needs lower precedence than everything else; assignment has precedence`
`-	// -1, everything else defined in mu::Parser has non-negative precedence`
`-	m_parser.DefineOprt(_T(";"), statementSeparator, -2);`
`+    m_parser.DefineOprt(_T(";"), statementSeparator, 0);`

Thus, to avoid a compiler warning on `unsigned a_iPri`, operator `;` cannot have negative precedence.
That makes current HEAD unable to parse/execute expressions that have both `=` and `;`.
For example executing `a=3; b=4; a*b` in a Note window raises the error:
"Unexpected operator '=' found at position -1"
The intended result can be obtained with `a=3, b=4, a*b`

gbm19:
> Could you please point me in the direction of the bug report that removing the ';' operator from muparser solves?

There is none. The bug was introduced in 62d4e5f and that code has not been released yet.

> Also could you at least mark the bugs fixed as 'fixed' or even 'closed' if you're really confident that they've been addressed. Otherwise, we'll be in a situation in a year or two's time having to test all the tickets to see whether they've been resolved.

For that I will need to have permissions to modify sf tickets

libscidavis/src/Graph.cpp
```diff
@@ -5332,8 +5332,8 @@ void Graph::updateCurveNames(const QString& oldName, const QString& newName, boo
         if (it->rtti() != QwtPlotItem::Rtti_PlotCurve)
             continue;
 
-        DataCurve *c = (DataCurve *)it;
-        if (c->type() != Function && c->plotAssociation().contains(oldName))
+        DataCurve *c = dynamic_cast<DataCurve *>(it);
+        if (c && c->plotAssociation().contains(oldName))
```
highperformancecoder, line 5336: Good. Now we can remove the check on line 5332 as well :)

libscidavis/src/MuParserScript.cpp
highperformancecoder: Thanks for the explanation.

Suthiro: Starting from [62d4e5f](https://github.com/highperformancecoder/scidavis/commit/62d4e5f64ed5e79b7eef3d11262f391024093c78) it is not possible to load a file, saved either using current HEAD or binary 2.3.0, that contains polynomial fit (and I believe any other muParser script). This commit does not make things better, however, changes the error message. Current error is 'Unexpected token "; a1=0.07..." found at position 20.' I can't find where from semicolons come, looks like it happens inside muParser. The attached [muParserTest.zip](https://github.com/highperformancecoder/scidavis/files/5861905/muParserTest.zip) is an example of non-loading file.
I kindly ask @gbm19 to see what could be done.

ChinouneMehdi: https://github.com/highperformancecoder/scidavis/commit/62d4e5f64ed5e79b7eef3d11262f391024093c78#diff-b1f8a9b558ed4ba1d51bb9a3571d6db9858dc473c9bb7381302057ff27512d5eL158

gbm19: Attached patch should solve the problem.

[additional-fix-for-PR153.patch.txt](https://github.com/highperformancecoder/scidavis/files/5862539/additional-fix-for-PR153.patch.txt)

PR #157, opened 2020-12-07
Group separator copy/paste bug

Suthiro:
Hello!
There is a bug connected with selection copy from a table with group separator option enabled.

Steps to reproduce:
1. Create an empty project
2. Make sure "group separators" option is enabled in preferences
3. Type in (or paste) any number long enough to be group-separated, e.g. 1234567890
4. Copy it whatever way
5. Paste and get the number split into several columns, e.g. 1	234	567	890.
or
6. Paste the result in your favorite text editor, e.g. gedit, also get number split into 1	234	567	890.
This behavior does not depend on format settings of the origin cell.

The pull request fixes the bug by using temporary locale when copying. I'm not sure if it is good to create a copy of locale each time, but also I'm not sure how to store such locale once and reuse in terms of current code organization.
Also, sorry for misindentation in the code, in CodeBlocks it looked good.

### DIFF
--- a/libscidavis/src/future/table/future_Table.cpp
+++ b/libscidavis/src/future/table/future_Table.cpp
@@ -31,7 +31,6 @@
 #include "table/future_Table.h"
 #include "core/Project.h"
 #include "lib/ActionManager.h"
-
 #include <QItemSelectionModel>
 #include <QTime>
 #include <QtGlobal>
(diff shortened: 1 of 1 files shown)


highperformancecoder:
Thanks for this. I agree - there is no sense in storing the locale, as next time it might be used, it might have changed (locale can be changed in preferences).

PR #161, opened 2020-12-21
Added an option to allow one to use either '.' or ',' as decimal separator on numerical input.

Suthiro:
This PR is a feature proposal.
Many applications ignore machine locale and use inbuilt '.' (dot) decimal separator. Also, sometimes data comes from outer world and also contain 'foreign' decimal separator, either '.' or ',' (comma). It is very inconvenient to replace separators externally before putting them into table, and many applications (for example, gnuplot) allows one to use both separators simultaneously. Of course, this should not be mandatory, so I added this as an option under "Numerical format".
What is changed exactly:
With the option checked, one is able to input both "1.23" and "1,23" as valid numbers for "English" (1.234) and "French" (1,234) types of locale. For "German" (1.000,123) and possibly some other formats with multiple or non '.'/',' separators it will make the numerical input impossible/wrong. I don't know how to handle it and even if it is to be handled.
What is not conformant with project coding style:
I decided not to add a boolean variable to `ApplicationWindow ` to keep settings, writing to/reading from `QtSettings` directly instead. I see no point to duplicate variables.
Also fixed a small (possible) bug in `ColumnPrivate.cpp`.

### DIFF
--- a/libscidavis/src/future/core/datatypes/String2DoubleFilter.h
+++ b/libscidavis/src/future/core/datatypes/String2DoubleFilter.h
@@ -34,6 +34,7 @@
 #include "lib/XmlStreamReader.h"
 #include <QXmlStreamWriter>
 #include <QtDebug>
+#include <QSettings>
 
 //! Locale-aware conversion filter QString -> double.
 class String2DoubleFilter : public AbstractSimpleFilter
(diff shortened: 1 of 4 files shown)


libscidavis/src/future/core/datatypes/String2DoubleFilter.h
```diff
@@ -42,33 +43,35 @@ class String2DoubleFilter : public AbstractSimpleFilter
 
 	public:
 		String2DoubleFilter() : d_use_default_locale(true) {}
+		// I'm not sure what this locale is, I searched through the whole source
+		// code and found zero references to the next two functions.
 		void setNumericLocale(QLocale locale) { d_numeric_locale = locale; d_use_default_locale = false; }
 		void setNumericLocaleToDefault() { d_use_default_locale = true; }
```
highperformancecoder, line 49: If you comment these lines out and the code base still builds, it would be safe to remove them.

libscidavis/src/future/core/datatypes/String2DoubleFilter.h
```diff
@@ -42,33 +43,35 @@ class String2DoubleFilter : public AbstractSimpleFilter
...
 
+		// Since valueAt() and isInvalid() do the same thing except the &ok part,
+		// I see reason to make a helper function and call it appropriately.
+       	// Please see convertToDouble() and getLocale() below
```
highperformancecoder, line 53: I appreciate your clear comments vis-a-vis a pull request, however longer term, these sorts of comments are not that useful in the code. Probably best if the code reviewer asks question on what is not clear, and respond with this sort of information in code review/pull request comments. In this case, I'd recognise a common refactoring out of common code, aka DRYing the code, so probably wouldn't have questioned it :).

libscidavis/src/future/core/datatypes/String2DoubleFilter.h
highperformancecoder: Note d_numeric_locale cannot be set without this method, so if the method is not called, the private members can be removed also. Probably added as part of some feature that never got finished, but someone who is no longer around. I suppose the who might be answered by running svn blame on the original code repo, which is still at SourceForge.

highperformancecoder:
This looks OK, so I'm merging it. As you note, it has to be a user specified option, as English and German deploy opposite values for decimal and thousands separators (. and , resp and vice versa). But a useful one nevertheless, for when it is known that the data does not use thousand separators (and you can't tell by inspection of the data). What is not clear from the code base is whether that user option is exposed on the GUI, or whether you have to edit the rather hard-to-find scidavis configuration file. If only the latter, I suspect this won't be used much.

libscidavis/src/future/core/datatypes/String2DoubleFilter.h
Suthiro: As I understand, first time `setNumericLocale` appears in [r1076](https://sourceforge.net/p/scidavis/svn/1076/) and used for ASCII import. In current revision `AsciiTableImportFilter` does not utilize this functionality.
Tried to comment out, nothing changed, code builds & runs. I will add this change in the next PR.

libscidavis/src/future/core/datatypes/String2DoubleFilter.h
Suthiro: TBH I don't like the current codebase: it is hard to understand what and where is going on, and lack of in-code comments make things worse. I agree about unnecessary verbosity in long-term, I would refrain from long comments in the future. Probably I should add these comments as a review to my own PR (not sure if this is possible).

Suthiro:
The option is exposed on the GUI.  I made changes to the `ConfigDialog`.
GUI

PR #162, opened 2020-12-22
Bug with pasting valid numerical data into numerical columns

Suthiro:
There is a bug with pasting valid numerical data into numerical columns when the locale is not "English" (1.000)

Steps to reproduce the bug:
1. Choose any locale format except (1.000) in the settings
2. Clipboard should contain any valid number with decimal separator (e.g. 1,234) or just copy any cell with such value.
3. Paste the cell whatever way in a numerical column
The column is now set to "Text" type despite having correct numerical value in the cell.

The `future_Table` uses it's own way to validate the input based on a regex. However, numerical column has an input filter for the same reason. The regex does not necessarily match the locale-aware and other settings-aware input filter of a column. Therefore, I exposed input filter of a column to validate an arbitrary string when data is pasted.

Also, I added a possibility to control the behavior when non-numeric data is pasted in to a numerical column: either ignore (set corresponding cells to invalid) or paste as is and convert whole column to text. The option is available through GUI and is on by default (i.e. converting whole column).

### DIFF
--- a/libscidavis/src/future/table/future_Table.cpp
+++ b/libscidavis/src/future/table/future_Table.cpp
@@ -51,6 +51,7 @@
 #include <QToolBar>
 #include <QtDebug>
 #include <QMimeData>
+#include <QSettings>
 #if QT_VERSION>=QT_VERSION_CHECK(5,10,0)
 #include <QRandomGenerator>
 #endif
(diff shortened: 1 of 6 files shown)


libscidavis/src/future/table/future_Table.cpp
```diff
@@ -509,18 +510,31 @@ void Table::pasteIntoSelection()
...
+        #else
+            QSettings settings(QSettings::NativeFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
+        #endif
+            bool convertToTextColumn = settings.value("/General/SetColumnTypeToTextOnInvalidInput", true).toBool();
+
+            for (int c=0; c<cols && c<input_col_count; c++)
+            {
+                Column * col_ptr = d_table_private.column(first_col + c);
+                if (convertToTextColumn)
+                if (col_ptr->columnMode() == SciDAVis::Numeric)
```
highperformancecoder, line 524: These two if statements should be combined into one, using an && conjoiner.

highperformancecoder:
I think I will need to switch SciDaVis master to using Ubuntu 18.04 for CI. I don't think we'll ever have SciDaVis 2.x compilable on OpenSUSE Tumbleweed, or other more modern Linuces...

alxpl:
> I think I will need to switch SciDaVis master to using Ubuntu 18.04 for CI. I don't think we'll ever have SciDaVis 2.x compilable on OpenSUSE Tumbleweed, or other more modern Linuces...

For what it's worth, even with gcc 11 that has broken a lot of things, version 2.3.0 still [builds fine in fedora rawhide](https://koschei.fedoraproject.org/package/scidavis?collection=f34).

PR #164, opened 2020-12-28
Bugfix in non-ASCII Origin file imports

Suthiro:
When one imports an `.opj` file containing non-latin characters anywhere (project name, labels, column descriptions, etc.), these are imported as "�" symbols (tested in Ubuntu 20.04 and Win 7). As I understand, `liborigin` imports strings as-is (i.e. byte sequence). However, I found that `.opj` files created in Windows with Russian locale contain strings encoded in [Windows-1251](https://en.wikipedia.org/wiki/Windows-1251) codepage. Therefore, it is a bug not to treat this properly during import. As always, I made the enhancement optional, so one can choose what codepage to use during import:
Codepage choice
Most users will use `US-ASCII` codepage which is default. The last used locale is stored in settings and also used to open files from `Recent Projects` submenu. This bugfix, of course, needs testing, but from my point of view, everything is ok.

Also, `QSettings` in the project is now accessed from one place and there is no need to repeat macro every time.

### DIFF
--- a/libscidavis/src/ApplicationWindow.cpp
+++ b/libscidavis/src/ApplicationWindow.cpp
@@ -125,7 +125,6 @@
 #include <QMdiSubWindow>
 #include <QTranslator>
 #include <QSplitter>
-#include <QSettings>
 #include <QApplication>
 #include <QMessageBox>
 #include <QPrinter>
@@ -3439,9 +3438,12 @@ void ApplicationWindow::open()
 {
 	OpenProjectDialog *open_dialog = new OpenProjectDialog(this, d_extended_open_dialog);
 	open_dialog->setDirectory(workingDir);
+	auto settings = getSettings();
+	open_dialog->setCodec(settings.value("/General/Dialogs/LastUsedOriginLocale", "").toString());
 	if (open_dialog->exec() != QDialog::Accepted || open_dialog->selectedFiles().isEmpty())
 		return;
 	workingDir = open_dialog->directory().path();
+    	settings.setValue("/General/Dialogs/LastUsedOriginLocale", open_dialog->codec());
 
 	switch(open_dialog->openMode()) {
 		case OpenProjectDialog::NewProject:
@@ -4231,8 +4233,7 @@ void ApplicationWindow::openTemplate()
 
 void ApplicationWindow::readSettings()
 {
-  QSettings settings;
-
+    auto settings = getSettings();
 	/* ---------------- group General --------------- */
 	settings.beginGroup("/General");
 #ifdef SEARCH_FOR_UPDATES
@@ -4508,13 +4509,7 @@ void ApplicationWindow::readSettings()
 
 void ApplicationWindow::saveSettings()
 {
-
-#ifdef Q_OS_MAC // Mac
-	QSettings settings(QSettings::IniFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
-#else
-	QSettings settings(QSettings::NativeFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
-#endif
-
+    auto settings = getSettings();
 	/* ---------------- group General --------------- */
 	settings.beginGroup("/General");
 #ifdef SEARCH_FOR_UPDATES
@@ -11791,6 +11786,7 @@ MultiLayer* ApplicationWindow::plotSpectrogram(Matrix *m, Graph::CurveType type)
 ApplicationWindow* ApplicationWindow::importOPJ(const QString& filename [[maybe_unused]])
 {
 #ifdef ORIGIN_IMPORT
+    auto codec = getSettings().value("/General/Dialogs/LastUsedOriginLocale", "").toString();
     if (filename.endsWith(".opj", Qt::CaseInsensitive) || filename.endsWith(".ogg", Qt::CaseInsensitive) || filename.endsWith(".org", Qt::CaseInsensitive))
     {
         QApplication::setOverrideCursor(QCursor(Qt::WaitCursor));
@@ -11804,14 +11800,14 @@ ApplicationWindow* ApplicationWindow::importOPJ(const QString& filename [[maybe_
         app->recentProjects.push_front(filename);
         app->updateRecentProjectsList();
 
-        ImportOPJ(app, filename);
+        ImportOPJ(app, filename, codec);
 
         QApplication::restoreOverrideCursor();
         return app;
     }
     else if (filename.endsWith(".ogm", Qt::CaseInsensitive) || filename.endsWith(".ogw", Qt::CaseInsensitive))
     {
-		ImportOPJ(this, filename);
+		ImportOPJ(this, filename, codec);
         recentProjects.removeAll(filename);
         recentProjects.push_front(filename);
         updateRecentProjectsList();
@@ -12397,7 +12393,10 @@ void ApplicationWindow::appendProject(const QString& fn)
         fn.contains(".ogw", Qt::CaseInsensitive) || fn.contains(".ogg", Qt::CaseInsensitive) ||
         fn.contains(".org", Qt::CaseInsensitive))
 #ifdef ORIGIN_IMPORT
-		ImportOPJ(this, fn);
+        {
+            auto codec = getSettings().value("/General/Dialogs/LastUsedOriginLocale", "").toString();
+            ImportOPJ(this, fn, codec);
+        }
 #else
 		{
                 QMessageBox::critical(this, tr("File opening error"),  tr("SciDAVis currently does not support Origin import. If you are interested in reviving and maintaining an Origin import filter, contact the developers.").arg(fn));
@@ -14155,4 +14154,13 @@ QStringList ApplicationWindow::tableWindows()
 	foreach(AbstractAspect *aspect, tables)
 		result.append(aspect->name());
 	return result;
+}
+
+QSettings ApplicationWindow::getSettings()
+{
+#ifdef Q_OS_MAC // Mac
+	return QSettings(QSettings::IniFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
+#else
+	return QSettings(QSettings::NativeFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
+#endif
 }
--- a/libscidavis/src/ApplicationWindow.cpp
+++ b/libscidavis/src/ApplicationWindow.cpp
@@ -3438,7 +3438,7 @@ void ApplicationWindow::open()
 {
 	OpenProjectDialog *open_dialog = new OpenProjectDialog(this, d_extended_open_dialog);
 	open_dialog->setDirectory(workingDir);
-	auto settings = getSettings();
+	auto& settings = getSettings();
 	open_dialog->setCodec(settings.value("/General/Dialogs/LastUsedOriginLocale", "").toString());
 	if (open_dialog->exec() != QDialog::Accepted || open_dialog->selectedFiles().isEmpty())
 		return;
@@ -4233,7 +4233,7 @@ void ApplicationWindow::openTemplate()
 
 void ApplicationWindow::readSettings()
 {
-    auto settings = getSettings();
+    auto& settings = getSettings();
 	/* ---------------- group General --------------- */
 	settings.beginGroup("/General");
 #ifdef SEARCH_FOR_UPDATES
@@ -4509,7 +4509,7 @@ void ApplicationWindow::readSettings()
 
 void ApplicationWindow::saveSettings()
 {
-    auto settings = getSettings();
+    auto& settings = getSettings();
 	/* ---------------- group General --------------- */
 	settings.beginGroup("/General");
 #ifdef SEARCH_FOR_UPDATES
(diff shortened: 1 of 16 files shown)


highperformancecoder:
This is failing CI because apparently QSettings is not copy constructible. Apparently it is not move constructible either, at least not on Ubuntu Bionic. See https://travis-ci.org/github/highperformancecoder/scidavis/builds/751797506 One possible solution is to have getSettings() return a unique pointer to a QSettings object, but there may be other possible solutions.

highperformancecoder:
> Most users will use `US-ASCII` codepage which is default. The last used locale is stored in settings and also used to open files from `Recent Projects` submenu. 

Why wouldn't most users use UTF-8? All other code pages are obsolete IIUC.

libscidavis/src/ApplicationWindow.cpp
```diff
@@ -14156,3 +14155,12 @@ QStringList ApplicationWindow::tableWindows()
...
 }
+
+QSettings ApplicationWindow::getSettings()
+{
+#ifdef Q_OS_MAC // Mac
+	return QSettings(QSettings::IniFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
+#else
+	return QSettings(QSettings::NativeFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
+#endif
+}
```
highperformancecoder: Could this be done as a Meyer singleton, or does the QSettings object have to be constructed each time it is used?

libscidavis/src/importOPJ.h
```diff
@@ -48,12 +49,15 @@ class ImportOPJ
...
 private:
+	bool setCodec(const QString& codecName);
+	QString decodeMbcs(char const * const input) const;
 	int translateOrigin2ScidavisLineStyle(int linestyle);
 	QString parseOriginText(const QString &str);
 	QString parseOriginTags(const QString &str);
 	int parse_error;
 	int xoffset;
 	ApplicationWindow *mw;
+	QTextCodec* d_codec;
```
highperformancecoder: Who owns d_codec? ISTM d_code will get leaked when the ImportOPJ object goes out of scope.

highperformancecoder: I just looked up the documentation of QTextCodec::codecForName. Apparently, the returned object is owned by the system, ownership is not passed. It is kind of an error in the API - that method should have returned a reference, not a pointer. Such things can be dealt with by some additional commentary - a comment to the effect that d_codec is a weak pointer to a system owned object would suffice here.

Suthiro:
> > Most users will use `US-ASCII` codepage which is default. The last used locale is stored in settings and also used to open files from `Recent Projects` submenu.
> 
> Why wouldn't most users use UTF-8? All other code pages are obsolete IIUC.

I'm not sure about latest builds of Origin, but older ones (at least including "2015" version) does use Windows' locale. So projects saved from Win 10 (apparenlty) use UTF-8, and from Win 7 and older are not. This option affects only import from Origin-generated files, and does not change anything else.

libscidavis/src/ApplicationWindow.cpp
Suthiro: You're right, singleton + pointer is better solution. I will do it after holidays.

libscidavis/src/importOPJ.h
Suthiro: See [QTextCodec::QTextCodec](https://doc.qt.io/qt-5/qtextcodec.html#QTextCodec-1). Since there is no `new QTextCodec()` call and the pointer is obtained by [QTextCodec::codecForName](https://doc.qt.io/qt-5/qtextcodec.html#codecForName), I suppose Qt owns codec and pointer to it and no leaks will occur.

Suthiro: Oh okay, my response is late, but you got the same conclusion. I will add some explanation to the code.

libscidavis/src/ApplicationWindow.cpp
Suthiro: After some considerations, I see two ways of implementing the singleton: a `QSharedPointer<QSettings>` + `static QSettings*` in each instance of `ApplicationWindow`, to cleanup settings after the last instance of `ApplicationWindow` goes out of scope; or create `QSettings` on stack/heap somewhere else as a separate unit and make it independent from `ApplicationWindow` completely. I'd prefer the latter approach, but I'd like to hear what do you think about it.

Also, `QSettings(const& QSettings)` is deleted indeed, but I wonder why everything compiles with `g++` on my machine _and_ with `msvcc` on the other.

highperformancecoder: IMHO, the only sort of singleton we should see in a C++ codebase is a [Meyer's singleton](http://laristra.github.io/flecsi/src/developer-guide/patterns/meyers_singleton.html).

Note Meyer's singletons should be initialised prior to any additional threads being created, to prevent race conditions.

highperformancecoder: Re compilation issues - maybe later versions of Qt define a move constructor for QSettings.

Suthiro: Okay, what do you think about:
```
QSettings& ApplicationWindow::getSettings()
{
        static QSettings d_settings(QSettings::NativeFormat,QSettings::UserScope, "SciDAVis", "SciDAVis");
	return d_settings;
}
```
I believe we will not have racing conditions, at least using modern compilers and C++11, according to [this answer on StackOverflow](https://stackoverflow.com/a/1661564). Do we need to support legacy compilers?

highperformancecoder: That is exactly what I had in mind. IFAIK, SciDaVis is not multithreaded, so there aren't any race conditions we need to consider.

It is nice to know that the Meyer's Singleton will be threadsafe in the next version of C++, and in practice is so with modern compilers. However, I would continue to ensure singletons are instantiated prior to any child threads being spawned - a sort of eager-lazy instantiation pattern as it were. It can be easily don by adding a call the Meyer's singleton within a static variable assignment in one of the .cpp files.

Suthiro:
Not sure what is wrong with tests now.

highperformancecoder:
This appears to be the problem:
readWriteProject.cpp:62:1: error: Failure in largeOriginImport: Unhandled exception: test crashed

See https://travis-ci.org/github/highperformancecoder/scidavis/pull_requests

I suspect Chinoue's Github action does not run the liborigin tests.

Suthiro:
Actually, looks like the test does not pass, and the problem lies inside my code. Small changes, so many troubles :) I will build with testing on current machine and investigate what's wrong.

ghost:
> This appears to be the problem:
> readWriteProject.cpp:62:1: error: Failure in largeOriginImport: Unhandled exception: test crashed
> 
> See https://travis-ci.org/github/highperformancecoder/scidavis/pull_requests
> 
> I suspect Chinoue's Github action does not run the liborigin tests.

Yes, It doesn't. Unless "-DORIGIN_IMPORT=ON" is passed to cmake.
I will push a fix for that.

Suthiro:
Do I understand right that `QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-runner'` indicates that under Ubuntu 18.04 tests are not configured properly? Looks like it fails because it simply is not able to find the file to open.

highperformancecoder:
The error is still readWriteProject.cpp:62:1: error: Failure in largeOriginImport: Unhandled exception: test crashed

But it is interesting that it only fails on Bionic Beaver, but passes on other distros such as Focal Fossa.

Suthiro:
> The error is still readWriteProject.cpp:62:1: error: Failure in largeOriginImport: Unhandled exception: test crashed
> 
> But it is interesting that it only fails on Bionic Beaver, but passes on other distros such as Focal Fossa.

Aside from my previous suggestion, I can only see the difference in `Qt` versions: it is 5.9.5 for 18.04, 5.12.8 for 20.04, 5.12.10/5.15.2 for MSVC and 5.15.2 for MinGW. However, I checked the docs and saw no difference concerning functions from  `QtCodec` and `QSettings` used in this PR.

Also I don't understand multiple `Couldn't load icon ` lines in 18.04 tests.

highperformancecoder:
> > The error is still readWriteProject.cpp:62:1: error: Failure in largeOriginImport: Unhandled exception: test crashed
> > But it is interesting that it only fails on Bionic Beaver, but passes on other distros such as Focal Fossa.
> 
> Aside from my previous suggestion, I can only see the difference in `Qt` versions: it is 5.9.5 for 18.04, 5.12.8 for 20.04, 5.12.10/5.15.2 for MSVC and 5.15.2 for MinGW. However, I checked the docs and saw no difference concerning functions from `QtCodec` and `QSettings` used in this PR.

OK - I'll take a look. I have a Bionic install around here somewhere...

> 
> Also I don't understand multiple `Couldn't load icon ` lines in 18.04 tests.

I don't understand those either, but they seem to be harmless.

Suthiro:
> OK - I'll take a look. I have a Bionic install around here somewhere...

Tried just now, the output is the same, but everything is fine from GUI.
EDIT: Now I don't get any errors and even "Couldn't load icon" lines for any configuration. I wonder what has changed system-wide.

Suthiro:
Found the bug and fixed.
I remember that I wrote the `nullptr`-initialization for `d_codec`, but somehow lost it later.
I'm very sorry.

P.S. I wonder why there is a difference between `qt` 5.9 and `qt` 5.12. They must fail _both_.

ghost:
Hi @Suthiro, I think you are new to git.
To push a pull-request to a repository It is recommended to fork another branch and then push it to the upstream/master.
If there are some required modifications amend them to the previous commit of that branch and then force-push it.
No need for all those commits.

Suthiro:
> Hi @Suthiro, I think you are new to git.
> To push a pull-request to a repository It is recommended to fork another branch and then push it to the upstream/master.
> If there are some required modifications amend them to the previous commit of that branch and then force-push it.
> No need for all those commits.

Hi @ChinouneMehdi! You're right, I'm new to _public_ git. The last push/force push were made by accident - forgot to switch to development branch, and I will be more careful later. Also you're right about development branches, it's a bad habit to work in master. I'm not sure about amending and force-pushing on a regular basis though, since I've read this is a troublemaking practice (see e.g. [this Q & A on stackoverflow](https://stackoverflow.com/a/432518)).
However, if @highperformancecoder will require/recommend to do so, I will comply with without doubt.

libscidavis/src/OpenProjectDialog.cpp
```diff
@@ -57,16 +58,31 @@ OpenProjectDialog::OpenProjectDialog(QWidget *parent, bool extended, Qt::WindowF
...
+    QHBoxLayout *openAs_layout = new QHBoxLayout();
+    openAs_layout->addWidget(new QLabel(tr("Open As")));
 	d_open_mode = new QComboBox();
 	// Important: Keep this is sync with enum OpenMode.
 	d_open_mode->addItem(tr("New Project Window"));
 	d_open_mode->addItem(tr("New Folder"));
-	advanced_layout->addWidget(d_open_mode);
-	setExtensionWidget(advanced_options);
+    openAs_layout->addWidget(d_open_mode);
+	advanced_layout->addLayout(openAs_layout);
```
highperformancecoder, line 73: Would much prefer this line appears directly after line 66, so as ownership transfer is obvious.

libscidavis/src/OpenProjectDialog.cpp
```diff
@@ -57,16 +58,31 @@ OpenProjectDialog::OpenProjectDialog(QWidget *parent, bool extended, Qt::WindowF
...
+    // Open As options
+    QHBoxLayout *openAs_layout = new QHBoxLayout();
+    openAs_layout->addWidget(new QLabel(tr("Open As")));
 	d_open_mode = new QComboBox();
 	// Important: Keep this is sync with enum OpenMode.
 	d_open_mode->addItem(tr("New Project Window"));
 	d_open_mode->addItem(tr("New Folder"));
-	advanced_layout->addWidget(d_open_mode);
-	setExtensionWidget(advanced_options);
+    openAs_layout->addWidget(d_open_mode);
```
highperformancecoder, line 72: This line should appear after line 68. See below.

libscidavis/src/OpenProjectDialog.cpp
highperformancecoder: Or use unique_ptr on line 66, and .release() here.

highperformancecoder:
On Fri, Jan 08, 2021 at 09:50:00PM -0800, Suthiro wrote:
>     Hi @Suthiro, I think you are new to git.
>     To push a pull-request to a repository It is recommended to fork another
>     branch and then push it to the upstream/master.
>     If there are some required modifications amend them to the previous commit
>     of that branch and then force-push it.
>     No need for all those commits.
> 
> Hi @ChinouneMehdi! You're right, I'm new to public git. The last push/force
> push were made by accident - forgot to switch to development branch, and I will
> be more careful later. Also you're right about development branches, it's a bad
> habit to work in master. I'm not sure about amending and force-pushing on a
> regular basis though, since I've read this is a troublemaking practice (see
> e.g. this Q & A on stackoverflow).
> However, if @highperformancecoder will require/recommend to do so, I will
> comply with without doubt.

Some projects require pull requests to have their individual commits
squashed. I don't for my projects - generally speaking I've found it
useful retaining the individual commits and messages, and for git
bisect, having smaller commits is desirable. That said, if you're
doing a ton of trivial commits, "fighting the CI system", then maybe
squashing them might be desirable.


-- 

                      http://www.hpcoders.com.au

PR #165, opened 2021-01-01
Fix MSVC build with latest MuParser (2.3.2)

ghost:


highperformancecoder:
Thanks. Largely equivalent to #undef min/#undef max that I've seen in other code bases. I don't know why MS continue to define these things as macros...

PR #170, opened 2021-01-03
Use preinstalled gl2ps library when possible.

ghost:


highperformancecoder:
Hi Chinoune,
I've been trying to make use of you cmake build process for the MacOSX build, which thank you is a vast improvement on that ghastly qmake system we had before.
For Macs, I will need to build qwt and qwtplot3d from source code - ideally, we should find the right versions of these code and place them as git submodules into the 3rdparty directory. I know where to get qwt5-qt5, but I was wondering if you could point me to which qwtplot3d repo I should use? I've tried sintegrial's, zohann's and syntheticcpp's from github, all to no avail. zohann's is closest, I could fork his code and fix the problems, but maybe there's another version out there that works better.

ghost:
I don't know any, I downloaded yours from your opensuse repository.

gbm19:
In Fedora I use package qwtplot3d-qt5-devel obtained from [src.fedora](https://src.fedoraproject.org/rpms/qwtplot3d-qt5) which uses [qwtplot3d-qt4](https://gitlab.com/anto.trande/qwtplot3d-qt4).

highperformancecoder:
In that case, I might need to create my own fork on github, so that I
can add it as a submodule.

I like that Chinoue's changes mean that one can easily use system
installed packages for Linux builds, and then "all-in-one" (via git
submodules) installations for the other distributions.

On Mon, Jan 04, 2021 at 02:53:27AM -0800, gbm19 wrote:
> In Fedora I use package qwtplot3d-qt5-devel obtained from src.fedora which uses
> qwtplot3d-qt4.
> 
> —
> You are receiving this because you modified the open/close state.
> 

-- 

                      http://www.hpcoders.com.au

PR #171, opened 2021-01-03
GitHub Actions.

ghost:
No testing with MSVC and MinGW-w64.


ghost:
depends on #169 and #170

highperformancecoder:
I'm a little unsure of this change, mostly because we're already using Travis for CI, for this and other projects I manage. But, because I'm curious at how Github actions work, I'm willing to try this as an experiment. But longer term, I doubt we can support both CI systems concurrently, so we'll need to choose.

PR #174, opened 2021-01-06
Fix compilation when PYTHONHOME is defined.

ghost:
With Python3 Py_SetPythonHome() expect wchar_t*. char* is not allowed.
https://docs.python.org/3/c-api/init.html#c.Py_SetPythonHome
https://docs.python.org/3/c-api/sys.html#c.Py_DecodeLocale

### DIFF
--- a/libscidavis/src/PythonScripting.cpp
+++ b/libscidavis/src/PythonScripting.cpp
@@ -211,7 +211,7 @@ PythonScripting::PythonScripting(ApplicationWindow *parent, bool batch)
     // if we need to bundle Python libraries with the executable,
     // specify the library location here
 #ifdef PYTHONHOME
-    Py_SetPythonHome(const_cast<char*>(str(PYTHONHOME)));
+    Py_SetPythonHome(Py_DecodeLocale(str(PYTHONHOME),NULL));
 #endif
     //		PyEval_InitThreads ();
 #if PY_MAJOR_VERSION >= 3


libscidavis/src/PythonScripting.cpp
```diff
@@ -39,7 +39,7 @@
 #include <iostream>
 
 #define str(x) xstr(x)
-#define xstr(x) #x
+#define xstr(x) L##x
```
highperformancecoder: I don't understand this change. This just prepends an L to x without quoting x, leading to uncompilable tokens like L/usr/local/share/python.

If the intention is to allow UTF-16 filenames, then I would caution that. Most systems would pass in UTF-8.

ghost: When I tested the patch I passed PYTHONHOME quoted.

highperformancecoder: On the command line, -DPYTHONHOME="/usr/local" would lead to /usr/local being passed. You would need to do something like -DPYTHONHOME=\"/usr/local\", which is a bit inconvenient.

Also, if you're adding L for the purposes of Visual C++ compatibility, then perhaps use the T() macro. Mind you I just tried the following experiment:

~~~~
#include <iostream>
#include <boost/locale.hpp>
using namespace std;
using boost::locale::conv::utf_to_utf;

#define str(x) xstr(x)
#define xstr(x) L##x

main()
{
  cout << utf_to_utf<char>(str(PYTHONHOME)) << endl;
}
~~~~~

And compiled this with 
c++ -DPYTHONHOME=\"β-release\" test.cc

And ran it:

zen>a.out
β-release


So it appears the compiler does correctly convert a UTF-8 string to UTF-16 with the L operator.

ghost: I have pushed a new patch.

highperformancecoder:
Thanks - that will work, and is probably optimal, as it makes no assumptions about the string being UTF-8. Actually my objections to using L were moot, as the compiler seems to convert a UTF-8 string corrrectly, or maybe even takes locale into account, with "C" locale being UTF-8. But thanks for leaving the stringify operation in - it's easier not to have to pur literal quotes around paths.

PR #176, opened 2021-01-08
Enable Testing on Windows.

ghost:


test/3dplot.cpp
```diff
@@ -11,7 +11,11 @@
 #include <stdexcept>
 using namespace std;
 
+#ifdef _WIN32
+#include <UnitTest++.h>
```
highperformancecoder, line 15: Couldn't you just add -I/mingw64 to the compiler options? This change will probably break my MXE build that I use to release Windows executables.

test/applicationWindow.cpp
```diff
@@ -5,7 +5,11 @@
 #include <iostream>
 using namespace std;
 
+#ifdef _WIN32
+#include <UnitTest++.h>
+#else
```
highperformancecoder, line 10: ditto

test/3dplot.cpp
ghost: This is required for both MSVC and MinGW.
It is cmake that add automatically the include directories using pkgconfig files.

highperformancecoder: I did a trial build. You are right - cmake adds the path to UnitTest, whereever it may be - the only reason for the #ifdef is that we're still using qmake for TravisCI. I should do some digital extraction and switch the Travis script to using cmake, that way we can just have the  #include <UnitTest++.h> without any #ifdef.

PR #184, opened 2021-01-14
[NFC] Formatting source files according to Qt rules.

ghost:
Using clang-format and _clang-format file copied from https://code.qt.io/cgit/qt/qt5.git/tree/_clang-format
NFC: No functional changes.


highperformancecoder:
Seems harmless enough, but I'm unlikely to run clang-format  myself on all commits :).

ghost:
> Seems harmless enough, but I'm unlikely to run clang-format myself on all commits :).

If you use qt-creator you can configure it to run clang-format automatically every time you save a file.

PR #186, opened 2021-01-16
Cmake adjustments to MSVC-related compilation and a fix to muParser-related entries.

Suthiro:
Brief: no need to search for muParser files and no need to include MUPARSER_INCLUDE_DIR if SCRIPTING_MUPARSER is disabled.
Also set up project to launch in Visual Studio and fix compilation troubles for newer SDKs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,8 @@
 cmake_minimum_required( VERSION 3.16 )
 
+# Target platform is Windows 10
+set(CMAKE_SYSTEM_VERSION 10.0)
+
 project( scidavis
   VERSION 2.3.0
   DESCRIPTION "SciDAVis is a free application for Scientific Data Analysis and Visualization."
@@ -17,6 +20,14 @@ set( CMAKE_C_EXTENSIONS OFF )
 get_property( MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG )
 
 set( CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH} )
+
+if( MSVC )
+  if (CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION VERSION_LESS 10)
+    message(SEND_ERROR "Windows 10 SDK is not found! Please, install one or set a CMAKE_WINDOWS_KITS_10_DIR environment variable to an absolute path to look for Windows 10 SDKs. The directory is expected to contain Include/10.0.* directories.")
+  else()
+	message(STATUS "Use Windows 10 SDK version ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
+  endif()
+endif()
 
 find_package( Qt5
   COMPONENTS
@@ -40,8 +51,9 @@ if( MSVC )
   # /wd4251 Silent dll-related warnings
   # /wd4127 Silent conditional expression is constant (Qt headers)
   # /wd4310 Silent cast truncates constant value (muParser headers)
+  # /wd4996 strcpy is unsafe warnings, equiv to #define _SCL_SECURE_NO_WARNINGS
   set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /W4 /permissive- \
-       /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310" )
+	 /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310 /wd4996" )
   set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /permissive-" )
   if( CMAKE_CXX_COMPILER_ID STREQUAL "Clang" )
     set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-nonportable-include-path" )
@@ -144,4 +156,8 @@ if( WIN32 )
 else()
   install( FILES ${DOC_FILES} DESTINATION share/doc/scidavis )
   install( FILES manual/index.html DESTINATION share/doc/scidavis/manual )
+endif()
+
+if( MSVC )
+  set_property(DIRECTORY PROPERTY VS_STARTUP_PROJECT "scidavis")
 endif()


ChinouneMehdi:
Have tried to build scidavis with SCRIPTING_MUPARSER=OFF.

CMakeLists.txt
```diff
@@ -40,9 +40,18 @@ if( MSVC )
   # /wd4251 Silent dll-related warnings
   # /wd4127 Silent conditional expression is constant (Qt headers)
   # /wd4310 Silent cast truncates constant value (muParser headers)
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /W4 /permissive- \
-       /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310" )
-  set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /permissive-" )
+  # /wd4996 strcpy is unsafe warnings, equiv to #define _SCL_SECURE_NO_WARNINGS
+  if (MSVC_VERSION LESS 1910)
+  # it looks like the /permissive- must be unset for newer SDKs
+  # https://developercommunity.visualstudio.com/content/problem/185399/error-c2760-in-combaseapih-with-windows-sdk-81-and.html
```
ChinouneMehdi, line 46: There is a difference between MSVC compiler version and Windows SDK version.
What are you trying to do here?!
The code build successfully with the latest MSVC 16.8.3 and Windows SDK 10.19041?!

CMakeLists.txt
```diff
@@ -40,9 +40,18 @@ if( MSVC )
...
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /W4 /permissive- \
+         /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310 /wd4996" )
+    set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /permissive-" )
+  else() # and everything builds finely just with
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /W4 \
+         /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310 /wd4996" )
+    set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4" )
+  endif()
```
ghost: There is a difference between MSVC compiler version and Windows SDK version.
The code build successfully with the latest MSVC 16.8.3 and Windows SDK 10.19041?!
1- The post said it has a problem to compile a code with Windows 8.1 SDK which is old (not new)!
2- You are disabling /permissive- with the newer MSVC which is working fine in my local machine and github-actions!
3- You said "It looks like", Is that an assumption or you are sure that it doesn't work.
Please clarify.

CMakeLists.txt
```diff
@@ -107,15 +116,17 @@ find_package( ZLIB "1.2.11" REQUIRED )
...
+  find_library( MUPARSER_LIB
+    NAMES muparser
+    REQUIRED
+    )
+  find_path( MUPARSER_INCLUDE_DIR
+    NAMES "muParser.h"
+    REQUIRED
+    )
+  message( STATUS "Found muParser : ${MUPARSER_LIB} include: ${MUPARSER_INCLUDE_DIR}" )
+endif()
```
ghost: Have tried to build scidavis with SCRIPTING_MUPARSER=OFF with your modifications.
Because it fails.
Muparser is not an optional dependency.

Suthiro: Then why we have it as an option? We should either remove the possibility to set `SCRIPTING_MUPARSER = OFF` or keep it as optional everywhere.

CMakeLists.txt
Suthiro: >  /wd4996 strcpy is unsafe warnings, equiv to #define _SCL_SECURE_NO_WARNINGS

this supresses multiple C4996 warnings.
> if (MSVC_VERSION LESS 1910)

this checks version of the toolset, since CMake is not able to check (easily) the SDK version. CMake generated solution with `/permissive-` for v140 and v141 toolkits (VS 2017) does not compile:
`1>C:\Program Files (x86)\Windows Kits\8.1\Include\um\combaseapi.h(229): error C2760: синтаксическая ошибка: ожидался токен "спецификатор типа", а не "идентификатор" (компилируется исходный файл ..\scidavis\libscidavis\src\Graph3D.cpp)`

My message about "newer SDKs" is misleading, sorry. VS2017 on Win7 x64 defaults to 8.1 SDK. While you're right about the issue with SDK and not with toolset, I see a point in disabling the `/permissive-` flag, since everything compiles and works fine without.

CMakeLists.txt
ghost: Muparser is not used just for scripting.
It is used with many other interactive tasks: plotting, analysing...

CMakeLists.txt
ghost: Like I said MSVC version has nothing to do with Windows-SDK and Windows 8.1 is an old SDK. You can update your SDK to Windows 10 even if you are working on Windows 7 (which itself not supported by Microsoft) as Windows SDKs are backward compatible.

You have to detect SDK version before disabling "/permissive-" which is essential for someone who wants his code to be standard conforming (without extensions).

ghost: If you are using Visual Studio to build, you can use [CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION.html) to detect SDK version.

Suthiro: From link:
>  If no Windows 10 SDK is available this value will be empty.

Moreover, I do have handful of 10.x SDKs and it is still empty.

CMakeLists.txt
Suthiro: I can't argue with that, but again, then we should remove `SCRIPTING_MUPARSER` _option_ completely and make everything related to muParser obligatory.

CMakeLists.txt
ghost: It is up to @highperformancecoder to decide, but I don't think Windows 7 is supported.
https://docs.microsoft.com/en-us/cpp/windows/how-to-use-the-windows-10-sdk-in-a-windows-desktop-application?view=msvc-160

CMakeLists.txt
ghost: Why should we remove an option enabled by default?

highperformancecoder:
On Sat, Jan 16, 2021 at 05:25:43AM -0800, Suthiro wrote:
> 
> Then why we have it as an option? We should either remove the possibility to
> set SCRIPTING_MUPARSER = OFF or keep it as optional everywhere.


My understanding (from some code commits, or perhaos ticket comments)
was there was an attempt some years back to remove Muparser
altogether, in favour of using Python. IIUC, that was never completed
to the point of being able to switch off muparser. Obviously, removing
muparser altogether will require a major version bump. Also, since we
still have issues in building and linking against python on Windows,
so even if the SCRIPTING_MUPARSER = OFF option worked, we'd still need
muparser for Windows.

-- 

                      http://www.hpcoders.com.au

highperformancecoder:
On Sat, Jan 16, 2021 at 09:13:28AM -0800, مهدي شينون (Mehdi Chinoune) wrote:
> @ChinouneMehdi commented on this pull request.
> 
> 
> In CMakeLists.txt:
> 
> > +  # /wd4996 strcpy is unsafe warnings, equiv to #define _SCL_SECURE_NO_WARNINGS
> +  if (MSVC_VERSION LESS 1910)
> +  # it looks like the /permissive- must be unset for newer SDKs
> +  # https://developercommunity.visualstudio.com/content/problem/185399/error-c2760-in-combaseapih-with-windows-sdk-81-and.html
> +    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /W4 /permissive- \
> +         /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310 /wd4996" )
> +    set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4 /permissive-" )
> +  else() # and everything builds finely just with
> +    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc /W4 \
> +         /wd4456 /wd4457 /wd4458 /wd4251 /wd4127 /wd4310 /wd4996" )
> +    set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /W4" )
> +  endif()
> 
> It is up to @highperformancecoder to decide, but I don't think Windows 7 is
> supported.

Not supported, but we don't deliberately exclude it either. I don't
have a Windows 7 system to test on. I'm trying to create a 64 bit
build of MXE to support SciDaVis, but running into endless
problems. With a circa 3 hour build time to try different options, it
is, as you imagine, incredibly slow going. When I switched to 64 bits
on the Minsky project, I found that Windows 7 was no longer supported,
as reported by a user.

I must admit, if we can set up a Github action to build an full .msi
artifact, then I might be well tempted to abandon MXE altogether, as
it is a real PITA to upgrade.


-- 

                      http://www.hpcoders.com.au

ghost:
> Not supported, but we don't deliberately exclude it either. I don't have a Windows 7 system to test on. I'm trying to create a 64  bit build of MXE to support SciDaVis, but running into endless problems. With a circa 3 hour build time to try different options, it is, as you imagine, incredibly slow going. When I switched to 64 bits on the Minsky project, I found that Windows 7 was no longer supported, as reported by a user. I must admit, if we can set up a Github action to build an full .msi artifact, then I might be well tempted to abandon MXE altogether, as it is a real PITA to upgrade.

It is in my TODO list.


Look forward to it. For reference, makeMsi,sh is a bash script that creates a .msi, given a successful build. It makes use of the WiX installshield program. It could be tacked onto a Mingw build script to generate an installable artifact.

CMakeLists.txt
highperformancecoder: Gnarrgh! github failed to include my response...

highperformancecoder: Not supported, but we don't deliberately exclude it either. I don't
have a Windows 7 system to test on. I'm trying to create a 64 bit
build of MXE to support SciDaVis, but running into endless
problems. With a circa 3 hour build time to try different options, it
is, as you imagine, incredibly slow going. When I switched to 64 bits
on the Minsky project, I found that Windows 7 was no longer supported,
as reported by a user.

I must admit, if we can set up a Github action to build an full .msi
artifact, then I might be well tempted to abandon MXE altogether, as
it is a real PITA to upgrade.

CMakeLists.txt
highperformancecoder: > Then why we have it as an option? We should either remove the possibility to set `SCRIPTING_MUPARSER = OFF` or keep it as optional everywhere.

My understanding (from some code commits, or perhaos ticket comments)
was there was an attempt some years back to remove Muparser
altogether, in favour of using Python. IIUC, that was never completed
to the point of being able to switch off muparser. Obviously, removing
muparser altogether will require a major version bump. Also, since we
still have issues in building and linking against python on Windows,
so even if the SCRIPTING_MUPARSER = OFF option worked, we'd still need
muparser for Windows.

CMakeLists.txt
Suthiro: Okay, just modified CMakeLists to look for Windows 10.x SDK and produce an error if nothing found. I will commit the changes after the decision about muParser option is made.

CMakeLists.txt
Suthiro: The changes I made reflect the only thing: we either keep everything related to muParser in `if (SCRIPTING_MUPARSER)` blocks, or remove the conditional blocks completely (for now). What is the point to include `MUPARSER_INCLUDE_DIR` always, but link against `MUPARSER_LIB` conditionally (as currently in `/libscidavis/CMakelists`)? It just does not make sense.

CMakeLists.txt
ghost: Well, you can use CMAKE_SYSTEM_VERSION.

CMakeLists.txt
ghost: > The changes I made reflect the only thing: we either keep everything related to muParser in `if (SCRIPTING_MUPARSER)` blocks, or remove the conditional blocks completely (for now). What is the point to include `MUPARSER_INCLUDE_DIR` always, but link against `MUPARSER_LIB` conditionally (as currently in `/libscidavis/CMakelists`)? It just does not make sense.

both MUPARSER_LIB and MUPARSER_INCLUDE_DIR are always defined.
https://github.com/highperformancecoder/scidavis/blob/master/libscidavis/CMakeLists.txt#L333
https://github.com/highperformancecoder/scidavis/blob/master/libscidavis/CMakeLists.txt#L348

CMakeLists.txt
ghost: I have pushed a commit that should fix compilation on old Windows.
#188

Suthiro:
Pushed an amendment to keep muParser entries intact, and to force using Windows SDK with version 10.x or above.

highperformancecoder:
Hi Chinoune, the changes look harmless enough to me. Are you happy with it? I don't personally use VC++ to build SciDaVis, so don't really have a dig in the fight.

ghost:
> Hi Chinoune, the changes look harmless enough to me. Are you happy with it? I don't personally use VC++ to build SciDaVis, so don't really have a dog in the fight.

OK.

ghost:
@Suthiro could you squash those two commits, because the second commit replace the first.

PR #187, opened 2021-01-16
Bugfix for SF#396 and handful of other various changes and bugfixes.

Suthiro:
Primarily this contains a bug fix for ["Date and time" <-> "Numeric" conversion](https://sourceforge.net/p/scidavis/scidavis-bugs/396/).
Also: fixed `setView()` called twice on Table initialization (all signals connected twice, that's why comments were not showing and controls were always visible);
DateTime conversion to double rules are now saved/loaded to/from project file.


Suthiro:
I don't know why travis fails with `Failure to find: src/future/core/datatypes/Double2DateTimeFilter.cpp`, since the entry is removed from Cmakelists.

libscidavis/src/future/core/column/Column.cpp
```diff
@@ -81,23 +81,32 @@ Column::~Column()
 
 void Column::setColumnMode(SciDAVis::ColumnMode mode, AbstractFilter *conversion_filter)
 {
-    if (mode == columnMode())
-        return;
```
highperformancecoder, line 85: I do have a strong preference for this "early return" style, rather than what you changed it to.

libscidavis/src/future/core/column/Column.cpp
```diff
@@ -327,7 +336,13 @@ void Column::save(QXmlStreamWriter *writer) const
         }
         break;
 
-    case SciDAVis::TypeQDateTime:
+    case SciDAVis::TypeQDateTime: {
+        // this conversion is needed to store base class;
+        auto numericFilter =
+                new NumericDateTimeBaseFilter(*(d_column_private->getNumericDateTimeFilter()));
```
highperformancecoder: Is this object owned by anything? This is the same horrible Qt style that makes ownership unclear. My preference when inserting an object into the Qt object tree is to immediately follow up by a setParent(() call after the object's creation, or better still to use the SciDAVisObject::addQChild template that creates the object already attached to the Qt object tree and returns a reference. Unfortunately, SciDAVisObject hasn't been back ported into the master branch, it is found in the classdesc-python branch, which is intended for SciDAVis 3.x.

libscidavis/src/future/core/column/Column.cpp
```diff
@@ -443,6 +461,19 @@ bool Column::load(XmlStreamReader *reader)
     return !reader->error();
 }
 
+bool Column::XmlReadNumericDateTimeFilter(XmlStreamReader *reader)
```
highperformancecoder, line 465: Any particular reason why reader is a pointer rather than a reference?

libscidavis/src/future/core/column/Column.h
```diff
@@ -157,6 +157,8 @@ class Column : public AbstractColumn
     void clear() override;
     //! This must be called before the column is replaced by another
     void notifyReplacement(const AbstractColumn *replacement) override;
+    //! Return the converter filter (for double <-> datetime  conversion)
+    NumericDateTimeBaseFilter *numericDateTimeBaseFilter() const;
```
highperformancecoder, line 161: I believe this method could return a reference.

libscidavis/src/future/core/column/ColumnPrivate.cpp
```diff
@@ -201,208 +203,207 @@ Column::Private::~Private()
     } // switch(d_data_type)
 }
 
-void Column::Private::setColumnMode(SciDAVis::ColumnMode mode, AbstractFilter *filter)
+void Column::Private::setColumnMode(SciDAVis::ColumnMode new_mode, AbstractFilter *converter)
 {
-    if (mode == d_column_mode)
+    const auto &old_mode = d_column_mode;
+    if (new_mode == old_mode)
```
highperformancecoder, line 209: Why this change?

libscidavis/src/future/core/column/ColumnPrivate.cpp
```diff
@@ -201,208 +203,207 @@ Column::Private::~Private()
...
     void *old_data = d_data;
     // remark: the deletion of the old data will be done in the dtor of a command
 
     AbstractSimpleFilter *new_in_filter, *new_out_filter;
-    bool filter_is_temporary {}; // it can also become outputFilter(), which we may not delete here
-    Column *temp_col = 0;
-    if (filter)
+    bool filter_is_temporary{
+        true
+    }; // it can also become outputFilter(), which we may not delete here
```
highperformancecoder: Please make the decl one line.

libscidavis/src/future/core/column/ColumnPrivate.cpp
```diff
@@ -201,208 +203,207 @@ Column::Private::~Private()
...
+    Column *temp_col = nullptr;
+
+    if (nullptr != converter)
         filter_is_temporary = false;
```
highperformancecoder, line 219: There are better ways of doing this. In this example, use a local unique_ptr, and a weak point combination:
auto converter=this->converter; // weak pointer
unique_ptr<...> local_converter;
...

if (!converter)
  ...
  local_convert.reset(new WhatEverConverter);
  converter=local_converter.reset();

...

No need for a flag variable. No need to delete these objects at return (allowing early return and exception safety). All upsides. At no stage is any object unowned.

this->converter should be converted to a unique_ptr also.

Ditto filter

libscidavis/src/future/core/column/ColumnPrivate.cpp
```diff
@@ -201,208 +203,207 @@ Column::Private::~Private()
...
-        } // switch(mode)
-        break;
-    }
-
-    // determine the new input and output filters
-    switch (mode) {
+    // prepare new d_data, d_data_type and filters
+    switch (new_mode) {
     case SciDAVis::Numeric: {
+        d_data = new QVector<double>();
```
highperformancecoder, line 225: Is d_data owned? If so, what happens to the old object pointed to by d_data?

libscidavis/src/future/core/column/ColumnPrivate.cpp
```diff
@@ -201,208 +203,207 @@ Column::Private::~Private()
...
+#endif
+    }
 
-    d_column_mode = mode;
+    // prepare temporary data
+    switch (old_mode) {
+    case SciDAVis::Numeric: {
+        disconnect(static_cast<Double2StringFilter *>(d_output_filter), SIGNAL(formatChanged()),
+                   d_owner, SLOT(notifyDisplayChange()));
+        temp_col = new Column("temp_col", *(static_cast<QVector<qreal> *>(old_data)), d_validity);
```
highperformancecoder: Make temp_col a unique_ptr.

libscidavis/src/future/core/datatypes/DateTime2DoubleFilter.h
```diff
@@ -1,66 +1,70 @@
...
+/***************************************************************************
+    File                 : DateTime2DoubleFilter.h
+    Project              : SciDAVis
+    --------------------------------------------------------------------
+    Copyright            : (C) 2007 by Tilman Benkert,
+                           Knut Franke
+    Email (use @ for *)  : thzs*gmx.net, knut.franke*gmx.de
+    Description          : Conversion filter QDateTime -> double (using Julian day).
+                           
 ***************************************************************************/
```
highperformancecoder: Spurious carriage return

libscidavis/src/future/core/datatypes/NumericDateTimeBaseFilter.h
```diff
@@ -0,0 +1,85 @@
...
+    NumericDateTimeBaseFilter(const UnitInterval unit = UnitInterval::Day,
+                              const QDateTime &date_time_0 = zeroOffsetDate)
+        : m_unit_interval(unit),
+          m_date_time_0((date_time_0.isValid())
+                                ? (date_time_0)
+                                : (zeroOffsetDate)){}; // default to zeroOffsetDate if invalid
+                                                       // datetime provided
+
+    NumericDateTimeBaseFilter(const NumericDateTimeBaseFilter &other)
+        : m_unit_interval(other.m_unit_interval), m_date_time_0(other.m_date_time_0){};
```
highperformancecoder, line 61: Can be declared =default

highperformancecoder:
A comment why I keep banging on about object ownership, RAII and exception safety. It's the only way to keep track of objects so that they get deleted at the appropriate time. The ideal is for no "delete" statements, and no naked "new" statements (they should only appear in smart pointer constructors or smart pointer reset statements). There is even an argument for banning "new" statements altogether - the replacements are make_shared and make_unique (C++17 onwards), but we needn't go quite that extreme yet. SciDAVisObject is a means of applying that philosophy to Qt object ownership. We do need to ban passing parent pointers on construction, that is very hard to reason, but a new followed by addParent or addChild is an acceptable alternative for now. 

Obviously, scidavis has a huge amount of "old style" non-RAII that needs to be refactored over time. Clearly this is not going to happen overnight, or possibly not even this decade, but we do need to be moving forward on this, so any new code should follow proper RAII guidelines, including using references rather than pointers on function interfaces wherever possible. It's called the "boy scout" principle.

libscidavis/src/future/core/column/Column.cpp
Suthiro: I will need another function then, because there is no way to handle the situation when the mode itself is not changed but parameters are. Is it okay?

libscidavis/src/future/core/column/Column.cpp
Suthiro: My bad! It is a temporary object therefore memory leak here. Fixed.

libscidavis/src/future/core/column/Column.cpp
Suthiro: Yes, every other similar function takes pointer, so I decided not to stand out from the crowd.

libscidavis/src/future/core/column/Column.h
Suthiro: The same as previous: other filters are returned by pointer.

libscidavis/src/future/core/column/ColumnPrivate.cpp
Suthiro: To get some clarity. There are many comparisons and switches below, it is much easier to comprehend what is compared to what. No problem to revert, but I'd prefer to keep meaningful variables' names.

libscidavis/src/future/core/column/ColumnPrivate.cpp
Suthiro: Did it, but FYI this conforms to clang format accepted recently. So it will happen again.

libscidavis/src/future/core/column/ColumnPrivate.cpp
Suthiro: Unfortunately, not in the current codebase. The converter is produced three (or four) functions above in the stack, and all this four functions will need to be rewritten to work with `QScopedPointer` (or `unique_ptr` or, better, references). I thought a bit about how this could be handled, and decided to keep it the way it was written before me.

libscidavis/src/future/core/column/ColumnPrivate.cpp
Suthiro: `d_data` is owned by class instance, for old `d_data` see line 211:

>     void *old_data = d_data;
>     // remark: the deletion of the old data will be done in the dtor of a command

It was before me. Again, I don't like this way of doing memory management, but I don't have time to rewrite huge amount of code and find all the bugs that will inevitably occur. Someday..

libscidavis/src/future/core/column/ColumnPrivate.cpp
Suthiro: Did it.

Suthiro:
I agree about RAII, this is why I used, for example, `QScopedPointer<NumericDateTimeBaseFilter>`. In my projects I almost never use raw `new`, but as I wrote above, there is too much legacy "old-style" code I don't have time to play with. If you are okay with inconsistency among similar functions, like `inputFilter()` and `outputFilter()` return `*Filter` and `numericalDateTimeFilter()` returns `&Filter`, I will use "modern style" as much as possible.
Also, I used `QScopedPointer` to stay inside Qt framework, but personally prefer `std::` counterparts. What is preferred for SciDAVis?

libscidavis/src/future/core/datatypes/DateTime2DoubleFilter.h
Suthiro: No idea where this came from.. Fixed

libscidavis/src/future/core/datatypes/NumericDateTimeBaseFilter.h
Suthiro: Declared default.

libscidavis/src/future/core/column/Column.cpp
highperformancecoder: > I agree about RAII, this is why I used, for example, `QScopedPointer<NumericDateTimeBaseFilter>`. In my projects I almost never use raw `new`, but as I wrote above, there is too much legacy "old-style" code I don't have time to play with. If you are okay with inconsistency among similar functions, like `inputFilter()` and `outputFilter()` return `*Filter` and `numericalDateTimeFilter()` returns `&Filter`, I will use "modern style" as much as possible.
> Also, I used `QScopedPointer` to stay inside Qt framework, but personally prefer `std::` counterparts. What is preferred for SciDAVis?

Good questions. I too prefer to use std:: counterparts, rather than Qt's, except when dealing with Qt's APIs (eg little point declaring a std::string if you're just going to pass it as a QString to a Qt method).

Re consistency, the horse has long bolted on that one :). I would say that for all new code, try to adhere to modern coding principles as  much as possible, use RAII, minimising the amount of raw pointers being passed around and so on. Hopefully, over time, with the "boy scout" principle, we'll get all of the code into shape. One of the first things we can do is grep for delete, and remove them by application of RAII. After that, we can focus on new, and replace the usage by SciDAVisObject, or smart pointer as appropriate.

libscidavis/src/future/core/column/Column.cpp
highperformancecoder: > Yes, every other similar function takes pointer, so I decided not to stand out from the crowd.

Fair enough - I'm just trying to encourage "boy scout" behaviour - improve the code incrementally by using "modern" style.

libscidavis/src/future/core/column/ColumnPrivate.cpp
highperformancecoder: OK fair enough.

libscidavis/src/future/core/column/ColumnPrivate.cpp
highperformancecoder: Gawd - that is ghastly! Well alright. I do use clang-format on another project that has strict style guide (that I have no control over), but personally I think coding style consistency just wastes too much time and energy, for very little benefit.

libscidavis/src/future/core/column/ColumnPrivate.cpp
highperformancecoder: OK well punt it for now, but this sort of crap does need to be addressed ultimately.

libscidavis/src/future/core/column/ColumnPrivate.cpp
highperformancecoder: Yes - I appreciate the time constraints :)!! You're right - urggh! I missed line 211. This stuff is impossible to reason about, the way it is written, but I agree beyond the scope of this PR.

libscidavis/src/future/core/column/Column.cpp
highperformancecoder: Fair enough.

highperformancecoder:
OK - I'll merge this once your push the changes you mentioned above.

libscidavis/src/future/core/column/ColumnPrivate.cpp
Suthiro: I don't mind this single-style stuff itself, since there is no manual effort. I wonder how to tweak our .clang file to get such declarations in one line.  Never used clang autoformat before, maybe, @ChinouneMehdi could help.

ghost: https://releases.llvm.org/11.0.0/tools/clang/docs/ClangFormatStyleOptions.html

Suthiro:
Do I need to squash & force push?

highperformancecoder:
> Do I need to squash & force push?

Up to you, I'm not too fussed either way.

Suthiro:
for some reason ` NumericDateTimeBaseFilter(const NumericDateTimeBaseFilter &other) = default;` makes cctor to be deleted. It is something with `Q_OBJECT`, so I reverted to previous (explicit) declaration.

highperformancecoder:
> for some reason ` NumericDateTimeBaseFilter(const NumericDateTimeBaseFilter &other) = default;` makes cctor to be deleted. It is something with `Q_OBJECT`, so I reverted to previous (explicit) declaration.

It inherits from AbstractAspect, which ought to be noncopiable, although I can't see where the copy operators have been deleted. This in turn inherits from QObject, which is non-copiable too, by design - I guess that is where the copy operation get deleted. This is not good - we need a shim class to insulate derived classes from non-copiable base classes. Maybe SciDaVisObject is a good place.

PR #188, opened 2021-01-17
Disable "/permissive-" with old Windows OSs.

ghost:


Suthiro:
No need doing this if Win7 is not supported. I made changes in #187 to keep `/permissive-` for MSVC builds by force requiring Win 10 SDK to be present.

PR #191, opened 2021-01-21
[GitHub-Actions] Checkout submodules on Windows.

ghost:
I have also removed fetch-depth as it is equal to 1 by default.


highperformancecoder:
Nice! Thanks for this.

PR #193, opened 2021-01-24
[NFC] PR for SF#368

Suthiro:
This PR addresses [ticket #368](https://sourceforge.net/p/scidavis/scidavis-bugs/368/). I consider it NFC but possibly buggy.

### DIFF
--- a/libscidavis/src/Graph.cpp
+++ b/libscidavis/src/Graph.cpp
@@ -130,7 +130,7 @@ Graph::Graph(QWidget *parent, QString name, Qt::WindowFlags f) : QWidget(parent,
 
     d_user_step = QVector<double>(QwtPlot::axisCnt);
     for (int i = 0; i < QwtPlot::axisCnt; i++) {
-        axisType << Numeric;
+        axisType << AxisType::Numeric;
         axesFormatInfo << QString();
         axesFormulas << QString();
         d_user_step[i] = 0.0;
@@ -369,14 +369,14 @@ void Graph::setAxesBaseline(QStringList &lst)
     }
 }
 
-QList<int> Graph::axesType()
+QList<Graph::AxisType> Graph::axesType()
 {
     return axisType;
 }
 
 void Graph::setLabelsNumericFormat(int axis, int format, int prec, const QString &formula)
 {
-    axisType[axis] = Numeric;
+    axisType[axis] = AxisType::Numeric;
     axesFormulas[axis] = formula;
 
     ScaleDraw *sd_old = (ScaleDraw *)d_plot->axisScaleDraw(axis);
@@ -409,7 +409,7 @@ void Graph::setLabelsNumericFormat(int axis, int format, int prec, const QString
 void Graph::setLabelsNumericFormat(int axis, const QStringList &l)
 {
     QwtScaleDraw *sd = d_plot->axisScaleDraw(axis);
-    if (!sd->hasComponent(QwtAbstractScaleDraw::Labels) || axisType[axis] != Numeric)
+    if (!sd->hasComponent(QwtAbstractScaleDraw::Labels) || axisType[axis] != AxisType::Numeric)
         return;
 
     int format = l[2 * axis].toInt();
@@ -427,10 +427,11 @@ QString Graph::saveAxesLabelsType()
 {
     QString s = "AxisType\t";
     for (int i = 0; i < 4; i++) {
-        int type = axisType[i];
-        s += QString::number(type);
-        if (type == Time || type == Date || type == DateTime || type == Txt || type == ColHeader
-            || type == Day || type == Month)
+        auto type = axisType[i];
+        s += QString::number(static_cast<int>(type));
+        if (type == AxisType::Time || type == AxisType::Date || type == AxisType::DateTime
+            || type == AxisType::Txt || type == AxisType::ColHeader || type == AxisType::Day
+            || type == AxisType::Month)
             s += ";" + axesFormatInfo[i];
         s += "\t";
     };
@@ -624,7 +625,7 @@ void Graph::changeTicksLength(int minLength, int majLength)
     emit modifiedGraph();
 }
 
-void Graph::showAxis(int axis, int type, const QString &formatInfo, Table *table, bool axisOn,
+void Graph::showAxis(int axis, AxisType type, const QString &formatInfo, Table *table, bool axisOn,
                      int majTicksType, int minTicksType, bool labelsOn, const QColor &c, int format,
                      int prec, int rotation, int baselineDist, const QString &formula,
                      const QColor &labelsColor)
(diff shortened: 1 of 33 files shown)


libscidavis/src/Graph.cpp
```diff
@@ -1828,10 +1829,10 @@ void Graph::updateCurvesData(Table *w, const QString &colName)
...
             }
             // compare setLabels*Format() calls in insertCurve()
-            if ((colType == Table::Text && atype != Txt)
-                || (colType == Table::Time && atype != Time)
-                || (colType == Table::Date && atype != Date)
-                || (colType == Table::DateTime && atype != DateTime))
+            if ((colType == SciDAVis::ColumnMode::Text && atype != AxisType::Txt)
+                || (colType == SciDAVis::ColumnMode::DateTime
+                    && ((atype != AxisType::Time) && (atype != AxisType::Date)
+                        && (atype != AxisType::DateTime))))
```
highperformancecoder, line 1835: Note: this code is not NFC. Could you please describe more why this change?

libscidavis/src/Table.cpp
```diff
@@ -1247,48 +1247,27 @@ QList<int> Table::plotDesignations()
...
-        case 4:
-            column(i)->setColumnMode(SciDAVis::Month);
-            break;
-        case 5:
-            column(i)->setColumnMode(SciDAVis::Day);
-            break;
-        }
-    }
+    for (int i = 0; i < d_future_table->columnCount(); i++)
+        column(i)->setColumnMode(ctl.at(i));
```
highperformancecoder, line 1270: This is also not an NFC change. I know this is part of an effort to get rid of Date and Time options in favour of DateTime, but required to support ancient SciDAVis project files.

highperformancecoder:
Hi Suthiro, I appreciate that the bulk of this change relates to using modern C++ class enums. I'm happy with those changes, even though I personally tend to write enums wrapped in structs:

struct Foo
{
    enum Type {bar, foobar, ...};
};

Then you can refer to Foo::bar etc - which solves the original leaky namespace problem, but has additional flexibility over class enums, you can inject the enumerators into another class's namespace via inheritence - eg class Bar: public Foo, for example.

Anyway, digression aside, could you please split out the two changes that change actual logic into a separate PR, such that this one is truly NFC?

libscidavis/src/Graph.cpp
Suthiro: Since `SciDAVis::ColumnMode` does not contain `Date` and `Time`, the comparison is the same for `Text` case, and for `SciDAVis::ColumnMode::DateTime` one is to check if axis type is not `Time`, `Date` and `DateTime` simultaneously. I think it is NFC. Why do you think it is not?

libscidavis/src/Table.cpp
Suthiro: The function now takes QList<SciDAVis::ColumnMode> as argument, so the switch is obsolete. Moreover, I just found that if this overload is commented out together with declaration, nothing changes.

Suthiro:
> Hi Suthiro, I appreciate that the bulk of this change relates to using modern C++ class enums. I'm happy with those changes, even though I personally tend to write enums wrapped in structs:
> 
> struct Foo
> {
> enum Type {bar, foobar, ...};
> };
> 
> Then you can refer to Foo::bar etc - which solves the original leaky namespace problem, but has additional flexibility over class enums, you can inject the enumerators into another class's namespace via inheritence - eg class Bar: public Foo, for example.

Hi, highperformancecoder!
I understand what do you mean, but it is misleading sometimes to have `Bar::Type` and `Baz::Type` while both is just the same. If an enum is shared across many classes, I feel it is better to keep the declaration separate and include if required.

> Anyway, digression aside, could you please split out the two changes that change actual logic into a separate PR, such that this one is truly NFC?

Let's decide if it is NFC or not, please, see my replies above. If not, I will try to separate changes.

libscidavis/src/Graph.cpp
highperformancecoder: The exact transformation will be (colType y.== Table::DateTime && atype != DateTime). If atype==Date or Time, then the code will follow different code paths. Of course this needs cleaning up, and your change is no doubt correct, but it would be good to focus on  the change on its own, rather than in conjunction with the other large amounts of change.

Possibly the way forward is to remove Date and Time from AxisType, and see what breaks?

NB I have long wanted to tidy these enums up, but have cowardly not done it, because of the appallingly low test coverage.

libscidavis/src/Table.cpp
highperformancecoder: It would be, if the data was constrained to the tags used in ColumnMode. But the data coming in is quite possibly from an old SciDAVis project file. It's not the way I would do schema evolution, but its that way it has been done :(.

Suthiro: Okay, I will revert this one to the original switch in this PR, despite this function is unused.

libscidavis/src/Graph.cpp
Suthiro: It is not possible to remove `Date` and `Time` from `AxisType` and keep it NFC (actually not break many thing in general), for example, in `Graph.cpp`
```
   line 735:  case Graph::AxisType::Month:
   line 745:  case Graph::AxisType::Time:
```
lead to different paths. I don't think we need to remove these, since they are used seemingly for a reason. Another question is about the comparison above. What do you think of:
```
            if ((colType == SciDAVis::ColumnMode::Text && atype != AxisType::Txt)
                || (colType == SciDAVis::ColumnMode::DateTime && atype != AxisType::Time) 
                || (colType == SciDAVis::ColumnMode::DateTime && atype != AxisType::Date)
                || (colType == SciDAVis::ColumnMode::DateTime && atype != AxisType::DateTime)))
```
While it is not strict NFC as I understand, it is very close and reflects changes clearer. This will not break anything, because `Time`, `Date` and `DateTime` curves use `QDateTime` under the hood.
Also I'm sure that it is not possible to keep this (quasi)intact and change everything else, I mean, it would not compile or would have bugs.

highperformancecoder: In the original code, if colType==DateTime && atype==Date, then this branch is taken. In this revised code, it is not. We need to establish under what circumstances atype is ever Date or Time, and what if anything, needs to be done about it, but I'd prefer a clean PR to look at these changes in isolation.

Suthiro: I see what you mean, but it is not possible to split this PR into two build-able PR's, because either we should add `Date` and `Time` to the `SciDAVis::ColumnMode` making this enum effectively the same as used previously in `Table::`, or accept this changes. Please note that `Table::Time` and `Table::Date` occurs exactly once in the whole project, here.

highperformancecoder:
OK - let just merge this, and hope there's no obscure bug created... We do need to clean these enums up.

PR #194, opened 2021-01-25
Update git submodules.

### LINKED ISSUE #40: Why does SciDAVis need pieces of liborigin installed on /usr/local?
Since SciDAVis uses a static liborigin library, I believe that the liborigin files in:
/usr/local/share/doc/liborigin/
/usr/local/include/liborigin/
and files:
/usr/local/lib/liborigin.*
/usr/local/bin/opj2dat
should not be installed on the system **during SciDAVis compilation**. It is not a good practice!
Of course, some of those files, like documentation, could be installed as part of the SciDAVis package, but not independently of it.
The liborigin.a file generated in 3rdparty/liborigin/ dir is not enough to build and run SciDAVis with Origin import support?

ghost:
- Include liborigin as a submodule.
- Update qwt5-qt5 and qwtplot3d.
- Change cmake files accordingly.
should fix #40


highperformancecoder:
I'm going to have to sort out Travis - maybe just delete .travis.yml

ghost:
I have pushed a fix for travis build.

.travis.yml
```diff
@@ -31,6 +31,8 @@ script:
   # required to avoid picking up a bodgy python interpreter
   - export PATH=/usr/bin:$PATH
   - export PYTHON=python3
+  - cp 3rdparty/liborigin.pro 3rdparty/liborigin/
+  - cp 3rdparty/config.h 3rdparty/liborigin/
```
highperformancecoder, line 35: You would also need to do a "git submodule update --init --recursive" within Travis.
But I think it better just to abandon Travis - github actions are working well.

ghost: No need for that, travis default to check submodules.
see: https://docs.travis-ci.com/user/customizing-the-build/#git-submodules

PR #197, opened 2021-01-26
Fix project configuration failure with Ninja generator on Windows.

ghost:


CMakeLists.txt
```diff
@@ -21,11 +18,15 @@ get_property( MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG )
 
 set( CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH} )
 
-if( MSVC )
+# Target platform is Windows 10
+if( CMAKE_GENERATOR STREQUAL "Visual Studio 15 2017" OR CMAKE_GENERATOR STREQUAL "Visual Studio 14 2015" )
+  set(CMAKE_SYSTEM_VERSION 10.0)
```
Suthiro: This line must precede the `project` line, otherwise it has no effect. See this [question](https://stackoverflow.com/a/45693321) on Stackoverflow and comment. Also from [CMAKE_SYSTEM_VERSION](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_VERSION.html) docs:

> .. CMAKE_SYSTEM_VERSION may be set explicitly when first configuring a new build tree ..

i.e. before `project`.
I tested this and put the line to the top by reason.

Suthiro (changes requested):
See comment below and please consider moving updated part before `project`.

PR #200, opened 2021-01-29
Build Windows installers.

ghost:


highperformancecoder:
I'm curious to see how this all works, as my alternative I've used for years requires a heavily convoluted bash script to generate the XML input that WIX requires to generate the even more convoluted database stashed in a .msi file. This looks much cleaner - but I'd like to replicate your approach on the Minsky project.

PR #205, opened 2021-02-05
Move to GoogleTest.

ghost:


highperformancecoder:
Could we please have a discussion at scidavis-contributors@lists.sf.net about this move? I personally have a weak preferance for UnitTest++, partly familiarity, partly because it is easier to run individual tests and partly because it is a bit better thought out - teg he test fixture can be the class you're testing, in gtest, it requires a whole bunch of dummy static methods to be defined. That said, I do use gtest on another code base I work on, gmock is a nice to have, as well as the ASSERT_* macros (which UnitTest++ doesn't have).

ghost:
> Could we please have a discussion at [scidavis-contributors@lists.sf.net](mailto:scidavis-contributors@lists.sf.net) about this move? I personally have a weak preferance for UnitTest++, partly familiarity, partly because it is easier to run individual tests and partly because it is a bit better thought out - teg he test fixture can be the class you're testing, in gtest, it requires a whole bunch of dummy static methods to be defined. That said, I do use gtest on another code base I work on, gmock is a nice to have, as well as the ASSERT_* macros (which UnitTest++ doesn't have).

Could you enable discussions in this repository where we can discuss things like that.
see: https://docs.github.com/en/discussions/quickstart

highperformancecoder:
On Fri, Feb 05, 2021 at 01:31:51PM -0800, مهدي شينون (Mehdi Chinoune) wrote:
>     Could we please have a discussion at scidavis-contributors@lists.sf.net
>     about this move? I personally have a weak preferance for UnitTest++, partly
>     familiarity, partly because it is easier to run individual tests and partly
>     because it is a bit better thought out - teg he test fixture can be the
>     class you're testing, in gtest, it requires a whole bunch of dummy static
>     methods to be defined. That said, I do use gtest on another code base I
>     work on, gmock is a nice to have, as well as the ASSERT_* macros (which
>     UnitTest++ doesn't have).
> 
> Could you enable discussions in this repository where we can discuss things
> like that.
> see: https://docs.github.com/en/discussions/quickstart

I'm not sure what you mean. Anyone can participate in this discussion
(attached to this pull request), but they won't if they don't know
about it.

In practice, there's only really 4 of us at the moment: Miquel,
Suthiro, you and me. There's about 2-3 other contributers who are kind
of quiet now. They should all be subscribed to
scidavis-contributors@lists.sf.net - and maybe the General Discussion
forum too.

I did something similar when I switched us from qtest to unittest++ a
couple of years ago - nobody had any strong opinions about what to use
- except for me, as qtest was driving me crazy.

I'm not opposed to switching to gtest, it has its pros and cons
relative to unitest++, I just thought it would be polite to at least
discuss this amongst the active developers.

Cheers

> 
> —
> You are receiving this because you commented.
> 

-- 

                      http://www.hpcoders.com.au

ghost:
The reasons why I want to switch to GoogleTest are:
- Popularity: GTest is more popular than UnitTest++ which means more tested. It is used by large projects like LLVM and Chromium.
- Active: See unittest-cpp/unittest-cpp#180 and unittest-cpp/unittest-cpp#183
- GTest has a better integration with development tools like CMake, Visual Studio and Qt Creator.
- As you can see in my PR, I didn't define any static methods, I have only created a new class:
```cpp
class ApplicationWindowTest : public ::testing::Test, public ApplicationWindow
{
};
```
https://github.com/highperformancecoder/scidavis/pull/205/files#diff-5df7a5b64ccec9873e9f1a49bb8026390b8671b92aa0f8671b3dd540034cf8acR8

test/CMakeLists.txt
```diff
@@ -23,39 +23,18 @@ if( ENABLE_Testing )
 
   add_executable( unittest ${SRCS} )
```
gbm19, line 24: You may consider renaming the executable to avoid confusion with UnitTest

ghost: What confusion?
We don't have another variable/option called "UnitTest".

gbm19: Since you are not using UnitTest++ anymore it seems to me that using 'unittest'  to name the test executable may sound misleading.

ghost: UnitTest++ and GoogleTest are both **[Unit testing](https://en.wikipedia.org/wiki/Unit_testing)** frameworks.
We still doing  **Unit testing**, nothing has changed.

Suthiro:
> They should all be subscribed to scidavis-contributors@lists.sf.net - and maybe the General Discussion forum too.

Oops, subscribed to the mailing list. 

> I'm not opposed to switching to gtest, it has its pros and cons relative to unitest++, I just thought it would be polite to at least discuss this amongst the active developers.

I also think it is good to discuss such things, thanks! However, in this particular case I don't mind the switch if the workflow isn't changed (at least much).

.github/workflows/main.yml
```diff
@@ -37,19 +37,26 @@ jobs:
...
+
       - name: Configuring
         run: |
           mkdir build && cd build
           cmake .. -DSEARCH_FOR_UPDATES=OFF -DDOWNLOAD_LINKS=OFF -DORIGIN_IMPORT=ON -DSCRIPTING_PYTHON=ON
 
       - name: Building
         run: |
-          make -j$(nproc) -C build
+          xvfb-run make -j$(nproc) -C build
```
highperformancecoder, line 59: Curious to know why it is necessary to run the make step under an X11 server?

test/CMakeLists.txt
highperformancecoder: Exactly :)

highperformancecoder:
Looks like everybody has had a chance to weigh in, so I accept the move to gtest.

.github/workflows/main.yml
ghost: > Curious to know why it is necessary to run the make step under an X11 server?

I don't know exactly.
When I pushed those changes the first time it failed with errors related to display. I think that gtest run some programs during configuration.

PR #209, opened 2021-03-09
bugfix for fit error messages in loop and subtle refactor

Suthiro:
Steps to reproduce the bug:
Choose data to fit, open "Fit Wizard"
Use custom invalid expression, like `a+scrt(x)` (the typo intended)
Try to fit and get 1+n error messages about an invalid parenthesis, where n is the number of datapoints.

Fixed this bug and also refactored some related code, mostly changed `double*` to `std::vector<double>` and made a function to generate x-vector, since the largest part of code in `calculateFitCurveData` in all the derived classes did the same thing.

Also, I'm not sure why MinGW test fails, since it fails for `master` in my repo and succeeds in upstream (this repo). See [upstream](https://github.com/highperformancecoder/scidavis/commit/83a36a6ceabc7b1ccffbfbf75e2ccf75947ae41b) and [fork](https://github.com/Suthiro/scidavis/commit/83a36a6ceabc7b1ccffbfbf75e2ccf75947ae41b) commits pointing to the same node `83a36a6ceabc7b1ccffbfbf75e2ccf75947ae41b` with different CI results. What?

### DIFF
--- a/libscidavis/src/ExponentialFit.cpp
+++ b/libscidavis/src/ExponentialFit.cpp
@@ -100,21 +100,13 @@ void ExponentialFit::storeCustomFitResults(const vector<double> &par)
         d_results[1] = 1.0 / d_results[1];
 }
 
-void ExponentialFit::calculateFitCurveData(const vector<double> &par, double *X, double *Y)
+bool ExponentialFit::calculateFitCurveData(const vector<double> &par, std::vector<double> &X,
+                                           std::vector<double> &Y)
 {
-    if (d_gen_function) {
-        double X0 = d_x[0];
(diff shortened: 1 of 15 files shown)


highperformancecoder:
The  MingW errors are warnings about missing strftime formatters - specifically %F and %T. You could replace these by combinations of more primitive formatters = eg "%F" = "%Y-%m-%d" and "%T"="%H:%M:%S". However, the error is actually in the libOrigin submodule, so in a sense, nothing to do with you :). I'm passing the PR.

Suthiro:
I thought the error is "70:  ModuleNotFoundError: No module named 'packaging'", but ok.

PR #211, opened 2021-03-16
Bugfix for SF#131

Suthiro:
Bugfix for [SF#131](https://sourceforge.net/p/scidavis/scidavis-bugs/131/): matrices processing are now fast upon undoing a formula.

NB: MSVC_Qt_5.12.10 builds successfully and MSVC_Qt_5.15.2 does not. The `PythonScripting.cpp` (is not changed in this PR) makes problems. I would like to fix this, but I don't know how. I would be very grateful for advice.

### DIFF
--- a/libscidavis/src/future/matrix/future_Matrix.cpp
+++ b/libscidavis/src/future/matrix/future_Matrix.cpp
@@ -1478,6 +1478,30 @@ void Matrix::transpose()
     RESET_CURSOR;
 }
 
+std::vector<std::vector<std::pair<double, bool>>>
+Matrix::getCells(const int startRow, const int endRow, const int startCol, const int endCol) const
+{
+    return d_matrix_private->getCells(startRow, endRow, startCol, endCol);
+}
+
+void Matrix::setCells(
+        const int startRow, const int startCol,
+                      const std::vector<std::vector<std::pair<double, bool>>> &values)
+{
+    WAIT_CURSOR;
+    auto lastRow = startRow + values.size() - 1;
+    auto lastCol = startCol
+            + std::max_element(values.cbegin(), values.cend(),
+                               [](const std::vector<std::pair<double, bool>> &a,
+                                  const std::vector<std::pair<double, bool>> &b) {
+                                   return a.size() > b.size();
+                               })
+                      ->size() - 1;
+    exec(new MatrixSetCellsCmd(d_matrix_private, startRow, lastRow, startCol, lastCol, values));
+    RESET_CURSOR;
+
+}
+
 void Matrix::mirrorHorizontally()
 {
     WAIT_CURSOR;
@@ -1779,6 +1803,48 @@ void Matrix::Private::setFormula(const QString &formula)
 {
     d_formula = formula;
     emit d_owner->formulaChanged();
+}
+
+std::vector<std::vector<std::pair<double, bool>>> Matrix::Private::getCells(const int startRow,
+                                                                            const int endRow,
+                                                                            const int startCol,
+                                                                            const int endCol) const
+{
+    if (startRow > endRow || startCol > endCol || startCol < 0 || endCol > columnCount()
+        || startRow < 0 || endRow > rowCount())
+        return std::vector<std::vector<std::pair<double, bool>>>();
+
+    std::pair<double, bool> invalidValue(std::numeric_limits<double>::quiet_NaN(), false);
+    std::vector<std::vector<std::pair<double, bool>>> values(
+            endRow - startRow + 1,
+            std::vector<std::pair<double, bool>>(endCol - startCol + 1, invalidValue));
+    for (int row = startRow; row <= endRow; row++)
+        for (int col = startCol; col <= endCol; col++)
+            values[row - startRow][col - startCol] = std::pair(cell(row, col), true);
+    return values;
+}
+
+void Matrix::Private::setCells(const int startRow, const int startCol,
+                               const std::vector<std::vector<std::pair<double, bool>>> &values)
+{
+    blockChangeSignals(true);
+    for (size_t ii = 0u; values.size() > ii; ++ii) {
+        const auto &column = values[ii];
+        for (size_t jj = 0u; column.size() > jj; ++jj) {
+            const auto &pair = column[jj];
+            if (!pair.second)
+                continue;
+            setCell(startRow + ii, startCol + jj, pair.first);
+        }
+    }
+    blockChangeSignals(false);
+    auto endRow = startRow + values.size() - 1;
+    auto endCol = startCol + std::max_element(
+            values.cbegin(), values.cend(),
+            [](const std::vector<std::pair<double, bool>> &a,
+               const std::vector<std::pair<double, bool>> &b) 
+			{ return a.size() > b.size();})->size() - 1;
+    emit d_owner->dataChanged(startRow, startCol, endRow, endCol);
 }
 
 } // namespace
--- a/libscidavis/src/future/matrix/matrixcommands.h
+++ b/libscidavis/src/future/matrix/matrixcommands.h
@@ -387,6 +387,42 @@ private:
 ///////////////////////////////////////////////////////////////////////////
 
 ///////////////////////////////////////////////////////////////////////////
+// class MatrixSetCellsCmd
+///////////////////////////////////////////////////////////////////////////
+//! Set cell values for multiple cells at once
+class MatrixSetCellsCmd : public QUndoCommand
+{
+public:
+    MatrixSetCellsCmd(future::Matrix::Private *private_obj, int first_row, int last_row,
+                      int first_column, int last_column,
+                      const std::vector<std::vector<std::pair<double, bool>>> &values,
+                      QUndoCommand *parent = 0);
+
+    void redo() override;
+    void undo() override;
+
+private:
+    //! The private object to modify
+    future::Matrix::Private * const d_private_obj;
+    //! The index of the first row
+    const int d_first_row;
+    //! The index of the last row
+    const int d_last_row;
+    //! The index of the first column
+    const int d_first_column;
+    //! The index of the last column
+    const int d_last_column;
+    //! New cell values
+    const std::vector<std::vector<std::pair<double, bool>>> d_values;
+    //! Backup of the changed values
+    const std::vector<std::vector<std::pair<double, bool>>> d_old_values;
+};
+
+///////////////////////////////////////////////////////////////////////////
+// end of class MatrixSetCellsCmd
+///////////////////////////////////////////////////////////////////////////
+
+///////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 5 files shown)


Suthiro:
I tried to check what has changed to make the build fail, and found that it is something SIP related, since it is not possible to build even previously buildable codebase: [this](https://github.com/Suthiro/scidavis/actions/runs/661398729) I pushed today and [this](https://github.com/highperformancecoder/scidavis/actions/runs/638353140) was pushed a week ago by highperformancecoder.

libscidavis/src/future/matrix/future_Matrix.cpp
```diff
@@ -1781,4 +1805,46 @@ void Matrix::Private::setFormula(const QString &formula)
...
+void Matrix::Private::setCells(const int startRow, const int startCol,
+                               const std::vector<std::vector<std::pair<double, bool>>> &values)
+{
+    blockChangeSignals(true);
+    for (size_t ii = 0u; values.size() > ii; ++ii) {
+        const auto &column = values[ii];
+        for (size_t jj = 0u; column.size() > jj; ++jj) {
+            const auto &pair = column[jj];
+            if (!pair.second)
+                continue;
```
highperformancecoder, line 1836: Just wondering about this. If the incoming matrix has no defined value for a particular element, then this will leave the old value unchanged. Might it be better to set the element to NAN?

libscidavis/src/future/matrix/matrixcommands.h
```diff
@@ -386,6 +386,42 @@ class MatrixSetRowCellsCmd : public QUndoCommand
...
+                      int first_column, int last_column,
+                      const std::vector<std::vector<std::pair<double, bool>>> &values,
+                      QUndoCommand *parent = 0);
+
+    void redo() override;
+    void undo() override;
+
+private:
+    //! The private object to modify
+    future::Matrix::Private * const d_private_obj;
```
highperformancecoder, line 406: OK - I understand that this should be a reference to the owning classes private implementation. We do need to move away from these ambiguous pointer types, but out of scope for this PR.

highperformancecoder:
> I tried to check what has changed to make the build fail, and found that it is something SIP related, since it is not possible to build even previously buildable codebase: [this](https://github.com/Suthiro/scidavis/actions/runs/661398729) I pushed today and [this](https://github.com/highperformancecoder/scidavis/actions/runs/638353140) was pushed a week ago by highperformancecoder.

Yup - there seems to be a sip error on Visual Studio. I'm going to have to let Chinoue sort of the VS stuff...

libscidavis/src/future/matrix/future_Matrix.cpp
Suthiro: See `Matrix.cpp:422`:
```
if (!isCellSelected(row, col))
                continue;
```

And `Matrix.cpp:431`:
```
if (!ret.isValid()) {
     return false;
}
```
If an invalid value encountered, the matrix will not be altered at all (compare to previous version of this function, when an invalid output from the script would left matrix in half-processed state). So only values of the selected cells are supposed to be altered, an this flag (`pair.second`) indicates if the cell should be modified.
I agree that variable name at `Matrix.cpp`:416 `std::pair<double, bool> invalidValue` is possibly misleading, since this not an invalid value, this is an `unsetValue`. I will change it in the next PR.

libscidavis/src/future/matrix/matrixcommands.h
Suthiro: There are many `Matrix->future::Matrix->future::Matrix::Private` chains of methods literally calling the very same method with the very same arguments. I wonder why so many levels of indirection are used, what `future` is supposed to mean and what was the plan?

highperformancecoder:
On Sun, Mar 21, 2021 at 03:09:42AM -0700, Suthiro wrote:
> @Suthiro commented on this pull request.
> 
> 
> In libscidavis/src/future/matrix/future_Matrix.cpp:
> 
> > +            if (!pair.second)
> +                continue;
> 
> See Matrix.cpp:422:
> 
> if (!isCellSelected(row, col))
>                 continue;
> 
> And Matrix.cpp:431:
> 
> if (!ret.isValid()) {
>      return false;
> }
> 
> If an invalid value encountered, the matrix will not be altered at all (compare
> to previous version of this function, when an invalid output from the script
> would left matrix in half-processed state). So only values of the selected
> cells are supposed to be altered, an this flag (pair.second) indicates if the
> cell should be modified.
> I agree that variable name at Matrix.cpp:416 std::pair<double, bool>
> invalidValue is possibly misleading, since this not an invalid value, this is
> an unsetValue. I will change it in the next PR.

Thanks for your explanation. So this is intended for use where a
selection of cells is pasted into the matrix, and we don't want to
change anything outside the selection.


-- 

Principal, High Performance Coders     ***@***.***
                      http://www.hpcoders.com.au

highperformancecoder:
On Sun, Mar 21, 2021 at 03:14:26AM -0700, Suthiro wrote:
> 
> There are many Matrix->future::Matrix->future::Matrix::Private chains of
> methods literally calling the very same method with the very same arguments. I
> wonder why so many levels of indirection are used, what future is supposed to
> mean and what was the plan?
> 

There was a rather abortionate attempt at implementing "aspect
oriented programming" that left us with all this half done "future"
classes. You can look up the history of that if interested.

Regardless of whether AOP as a software engineering technique has
merit, the implementation chosen here is way overengineered and
violates just about every C++ principle one can think of.

But its a big job just to remove and clean up all these classes, and
we have more urgent priorities, we'll have to live with them for now
:). I had a go at tracing out the table class, and effected a small
amount of cleanup, but that was days of work, and multiple A4 pages
just to draw the class diagram.


-- 

Principal, High Performance Coders     ***@***.***
                      http://www.hpcoders.com.au

libscidavis/src/future/matrix/matrixcommands.h
Suthiro: > On Sun, Mar 21, 2021 at 03:14:26AM -0700, Suthiro wrote: There are many Matrix->future::Matrix->future::Matrix::Private chains of methods literally calling the very same method with the very same arguments. I wonder why so many levels of indirection are used, what future is supposed to mean and what was the plan?
> There was a rather abortionate attempt at implementing "aspect oriented programming" that left us with all this half done "future" classes. You can look up the history of that if interested. Regardless of whether AOP as a software engineering technique has merit, the implementation chosen here is way overengineered and violates just about every C++ principle one can think of. But its a big job just to remove and clean up all these classes, and we have more urgent priorities, we'll have to live with them for now :). I had a go at tracing out the table class, and effected a small amount of cleanup, but that was days of work, and multiple A4 pages just to draw the class diagram.
> […](#)
> --
> ---------------------------------------------------------------------------- Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance Coders ***@***.*** http://www.hpcoders.com.au ----------------------------------------------------------------------------

Thanks for the clarification!
I found that sometimes it is very hard to comprehend the current code base. Initially, I thought it is due to my rather poor level of programming. As time passes, I see that my level is programming is poor indeed, but comprehension troubles connected with "violation just about every C++ principle one can think of" present also :)
I've read about AOP(D) and TBH don't like the idea. Maybe I'm just a layman.
As a conclusion: do we want to keep AOP paradigm? While I will not do anything with this right now, but it could be important later.

Suthiro:
> > I tried to check what has changed to make the build fail, and found that it is something SIP related, since it is not possible to build even previously buildable codebase: [this](https://github.com/Suthiro/scidavis/actions/runs/661398729) I pushed today and [this](https://github.com/highperformancecoder/scidavis/actions/runs/638353140) was pushed a week ago by highperformancecoder.
> 
> Yup - there seems to be a sip error on Visual Studio. I'm going to have to let Chinoue sort of the VS stuff...

I fear Chinoune's account has been removed recently. I don't know why and when.

highperformancecoder:
On Mon, Mar 22, 2021 at 06:28:30AM -0700, Suthiro wrote:
> 
> I fear Chinoune's account has been removed recently. I don't know why and when.

It's still there: https://github.com/MehdiChinoune. But he hasn't been
active on SciDAVis for the last month, and also appears to have
withdrawn from the SciDaVis organisation he set up. It looks like a
case of other priorities.


-- 

Principal, High Performance Coders     ***@***.***
                      http://www.hpcoders.com.au

highperformancecoder:
On Mon, Mar 22, 2021 at 06:25:09AM -0700, Suthiro wrote:
>     There was a rather abortionate attempt at implementing "aspect oriented
>     programming" that left us with all this half done "future" classes. You can
>     look up the history of that if interested. Regardless of whether AOP as a
>     software engineering technique has merit, the implementation chosen here is
>     way overengineered and violates just about every C++ principle one can
>     think of. But its a big job just to remove and clean up all these classes,
>     and we have more urgent priorities, we'll have to live with them for now
>     :). I had a go at tracing out the table class, and effected a small amount
>     of cleanup, but that was days of work, and multiple A4 pages just to draw
>     the class diagram.
>     …
> 
>     ----------------------------------------------------------------------------
>     Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance
>     Coders @.*** http://www.hpcoders.com.au
>     ----------------------------------------------------------------------------
> 
> Thanks for the clarification!
> I found that sometimes it is very hard to comprehend the current code base.
> Initially, I thought it is due to my rather poor level of programming. As time
> passes, I see that my level is programming is poor indeed, but comprehension
> troubles connected with "violation just about every C++ principle one can think
> of" present also :)
> I've read about AOP(D) and TBH don't like the idea. Maybe I'm just a layman.
> As a conclusion: do we want to keep AOP paradigm? While I will not do anything
> with this right now, but it could be important later.
> 

If I have any say over this, and I suppose I do, then "no way, no
how". The AOP work that has been done has been a complete disaster. If
anything, I'd like to introduce classdesc as a low code means of doing
similar sorts of things - I made a start with the classdesc-python
branch, but the big disadvantage has been a blowout in compile times,
something I need to get back to solving.

Cheers
-- 

Principal, High Performance Coders     ***@***.***
                      http://www.hpcoders.com.au

Suthiro:
> On Mon, Mar 22, 2021 at 06:28:30AM -0700, Suthiro wrote: I fear Chinoune's account has been removed recently. I don't know why and when.
> It's still there: https://github.com/MehdiChinoune. But he hasn't been active on SciDAVis for the last month, and also appears to have withdrawn from the SciDaVis organisation he set up. It looks like a case of other priorities.
> ---------------------------------------------------------------------------- Dr Russell Standish Phone 0425 253119 (mobile) Principal, High Performance Coders ***@***.*** http://www.hpcoders.com.au ----------------------------------------------------------------------------

Ah ok. He could not be mentioned from here using "@", his pull requests are now belong to "ghost" (and github says it is for removed accounts) and there is no mention in "contributors". Based on this I decided that the account was removed.

PR #215, opened 2021-04-16
Bugfix

Suthiro:
Found and fixed a bug with DateTime input filter. Refactored code a bit.
[NFC] Changed name for a variable as discussed earlier.
Restored translations & updated Russian.


libscidavis/src/future/table/TableView.cpp
```diff
@@ -728,9 +728,8 @@ void TableView::applyType()
                 break;
             }
             col->setColumnMode(new_mode, converter);
-            DateTime2StringFilter *filter =
-                    static_cast<DateTime2StringFilter *>(col->outputFilter());
-            filter->setFormat(format);
+            static_cast<DateTime2StringFilter*>(col->outputFilter())->setFormat(format);
+            static_cast<String2DateTimeFilter*>(col->inputFilter())->setFormat(format);
```
highperformancecoder: I would prefer dynamic_cast and testing for valid cast here, rather than static_cast. But you're following the existing poor coding practices, so OK.

highperformancecoder: Line 732 must be the actual bugfix, right?

highperformancecoder:
Any idea why xvfb-run is not starting on Ubuntu 20.04?

libscidavis/src/future/table/TableView.cpp
Suthiro: > I would prefer dynamic_cast and testing for valid cast here, rather than static_cast. But you're following the existing poor coding practices, so OK.

TBH I just have not thought about it. Will fix now. And you're not quite right about my "poor coding practices", I refactored these nasty `const char * []` loops :)

> Line 732 must be the actual bugfix, right?

Yes, line 732 is the bugfix.

Suthiro:
> Any idea why xvfb-run is not starting on Ubuntu 20.04?

I've encountered this bug recenlty, and it was enough to re-run jobs. It is something with either latest Ubuntu update(s), or github's deploying system. I hope it will not persist.

BTW, what exactly is needed to release new version?

libscidavis/src/future/table/TableView.cpp
highperformancecoder: Sorry if it sounded like I was criticising you - I wasn't, I was criticising the code quality of scidavis as we received, and this was just a suggestion of a small step we can take to improve it. Thanks for the update.

highperformancecoder:
> > Any idea why xvfb-run is not starting on Ubuntu 20.04?
> 
> I've encountered this bug recenlty, and it was enough to re-run jobs. It is something with either latest Ubuntu update(s), or github's deploying system. I hope it will not persist.

Seems like you're right - it passed this time. It's a bit worrying though, that the CI system is unreliable. I've got the same problem currently with Travis and docker images that don't run properly on another project, so I'm thinking of porting those over to Github actions...
> 
> BTW, what exactly is needed to release new version?

I need to solve why the MacOS bundle crashes on startup. I can't quite remember now, but there may be similar issues with the Windows build.

libscidavis/src/future/table/TableView.cpp
Suthiro: It's all right, just was surprised a bit :) But still you're right, so I will try to think about such things more thoroughly.

Suthiro:
> I need to solve why the MacOS bundle crashes on startup. I can't quite remember now, but there may be similar issues with the Windows build.

SIP problems, I suppose. We might need to remove sip bindings for Qwt related stuff.

PR #216, opened 2021-04-19
Follow up bugfix for PR#187.

Suthiro:
Found a bug introduced in #187.
Simple back and forth conversion returns false for validity of m_date_time_0 set to Julian day 0 (-4714-24-11 12:00:00:000) and true for something like 2001-01-01 00:00:00.000. 

```
QDateTime tTime = QDateTime::fromString(m_date_time_0.toString("yyyy-dd-MM hh:mm:ss:zzz"), 
     "yyyy-dd-MM hh:mm:ss:zzz");
bool t = tTime.isValid();
```

so I decided to change the default date to the start of unix epoch and store it in milliseconds.

### DIFF
--- a/libscidavis/src/future/core/datatypes/NumericDateTimeBaseFilter.cpp
+++ b/libscidavis/src/future/core/datatypes/NumericDateTimeBaseFilter.cpp
@@ -29,35 +29,37 @@
 #include "NumericDateTimeBaseFilter.h"
 #include <math.h>
 
-const QDateTime NumericDateTimeBaseFilter::zeroOffsetDate =
-        QDateTime(QDate::fromJulianDay(0), QTime(12, 0, 0, 0));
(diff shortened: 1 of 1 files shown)


highperformancecoder:
Hi Suthiro, trying to understand the impact of this. IIUC, this changes any datetimes stored in project files from ISO format to milliseconds from epoch? Wouldn't that be a schema breaking change?

Suthiro:
As far as I'm concerned, this affects only `NumericalDateTimeFilter` I introduced in #187, so it breaks the schema for all those that built scidavis themselves. I think this is not the most used feature, so I doubt it does really matter. The only thing I can think of is the inconsistency between `DateTime` storage format inside the `xml`: some will be kept in ISO format, and this one in miliseconds. If it is important, I will revert to ISO format but set "2000-01-01 00:00:00" as a default date.

highperformancecoder:
OK - I can accept that.

PR #219, opened 2021-05-11
Update Graph3D.cpp

alxpl:
Fix typos

### DIFF
--- a/libscidavis/src/Graph3D.cpp
+++ b/libscidavis/src/Graph3D.cpp
@@ -1945,7 +1945,7 @@ void Graph3D::print()
 
 void Graph3D::copyImage()
 {
-    QApplication::clipboard()->setPixmap(QPixmap::fromImage(sp->grabFramebuffer()),
+    QApplication::clipboard()->setPixmap(QPixmap::fromImage(sp->grabFrameBuffer()),
                                           QClipboard::Clipboard);
     sp->updateData();
 }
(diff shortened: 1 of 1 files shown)


highperformancecoder:
I guess this change needs to be propagated into the Qwt repo submoduled into scidavis. Can you create PRs dealing with that?

alxpl:
These three were reported by g++ during compilation (which stopped at that point), nothing else.

grep found two more instances:

[Line 175 of qwtplot3d/src/qwt3d_io_gl2ps.cpp ](https://github.com/SciDAVis/qwtplot3d/blob/438c855d8a02bdf6ba7a20af269f12ee8f74ebd2/src/qwt3d_io_gl2ps.cpp#L175)
and
[line 232 of qwtplot3d/src/qwt3d_io.cpp ](https://github.com/SciDAVis/qwtplot3d/blob/438c855d8a02bdf6ba7a20af269f12ee8f74ebd2/src/qwt3d_io.cpp#L232)

(Sorry for not submitting a PR, I'm away from the computer that's set up with github keys and everything)

Suthiro:
Just FYI these lines were changed recently in #207:
```
git blame libscidavis/src/Graph3D.cpp -L 1948,1949
> f5c4629bb (Mehdi Chinoune 2021-02-02 06:56:43 +0100 1948)
QApplication::clipboard()->setPixmap(QPixmap::fromImage(sp->grabFramebuffer()), QClipboard::Clipboard);
...
```
The qwt submodule was also affected.

